PR #311, opened 2015-02-27
First pass at adding an admittance matrix method

jgostick:
This method returns an admittance matrix, aka a laplacian matrix, which has 'almost' the same values and sparsity pattern as our alg.A matrix.  The only differences are due to the way we add boundary conditions.  I think this could be remedied if we switch to adding all BC with a 'source term' approach.  Not sure if this is faster than our custom built coefficient matrix or not.


jgostick:
After extensive testing, @maghighi found this no faster and needing lots of work to be useful...therefore abandon.

PR #332, opened 2015-04-15
Feature/workspace

jgostick:
This PR introduces the possibility of placing some of the Controller functionality at the package level, such as save and load.  

This means that as soon as OpenPNM is imported it creates a 'workspace' which contains ALL objects.  This 'workspace' can be saved using OpenPNM.save(), or a saved workspace can be loaded from a file.  It is an interesting development that could use a bit more thought and refinement to get right.


bspellio:
I like the idea of this functionality, but I don't think we should have it be the default. It would be great to have a command like `OpenPNM.Workspace.create()` which would then attach the Controller functionality to the package level.

jgostick:
This would also entail changing the name of controller to workspace.  I wonder if we could provide two options...(1) adds save/load/etc functionality to the package level (used for 99% of cases) and (2) return a handle to what we currently call the 'Controller', which has more hardcore functionality for power users.

bspellio:
@jgostick Let's rebase this branch, add a deprecation warning to using `Controller` directly, and then we can merge it.

jgostick:
It's not really a deprecation.  I prefer to think of these methods at the
package level as 'helper' methods.  Scipy is filled with them, so us
'engineers' are used to it.  For eg. there is sp.rand with is just a
shortcut to sp.random.rand.  It saves a whole 7 keystrokes!  The actual
point is not a time saver, but a conceptual organization.  People expect
save/load etc to be on the package level.

Go ahead and rebase it, but I still need to add all the other functionality
so don't merge yet.

On Thu, Jun 18, 2015 at 10:23 AM, Brennan Spellacy <notifications@github.com

bspellio:
@jgostick please close this PR and reopen it against `master`.

PR #343, opened 2015-05-05
Fixing all tests

bspellio:
This PR is addressing all of the errors present in the develop branch in addition to adding coveralls support and the use of tox for testing. To run the testing suite simply run 
```
$ tox
```
 This command will output the results of the test suite. To run the test suite and then print the corresponding coverage, run:
```
$ tox -- --cov OpenPNM
```


bspellio:
Please review @RodericDay

bspellio:
You may want to take a look to @jgostick so that you have an idea of what doctests we thought didn't bring much to the table.

bspellio:
I've added testing for python version 3.2 and 3.3 onto TravisCI, they removed support for 3.1 due to compatibility issues. As for installing scipy on windows @jgostick , it doesn't look like there is wheel support yet. The popular option is to get a C and fortran compiler on your laptop. The test suite can still be run locally with `py.test`, so if you decide to do that, everything should be fine. I only implemented tox so that we could test OpenPNM in a similar way that TravisCI will so that less time was wasted working on builds, but in the end, tox is really just a luxury because we have Travis to fall back on.

bspellio:
And after some more failures/reading it's come to my attention that conda only supports python 3.3+. This means that we can't support anything besides 3.3 and 3.4 with complete confidence. Personally, I don't think it's that bad since most people using python 3 are pretty good about updating their systems.

RodericDay:
:+1:

Issue #334, opened 2015-04-15
Implement Code Linting


bspellio:
Implementing code linting allows us to automate the process of ensuring that all the code we merge is consistent stylistically. This encourages quality in the code base and makes it more approachable. I believe code linting is going to be coming to code climate soon, but until then we should have a substitute. My recommendation would be that we use the PEP8 style guide, for the most part it's the industry standard: https://www.python.org/dev/peps/pep-0008/

jgostick:
The Spyder IDE that we all use supports the pylint plugin.  The only problem is that it gets very distracting when trying to code, because we get so many warnings.  :-D

bspellio:
Absolutely, that's why automation can be great. 
* Write you code and be aware of the style to the best of your ability
* Open a PR
* Make all the needed style changes in a single commit :+1:

PR #345, opened 2015-05-08
Pep8ification of OpenPNM

bspellio:
This PR addresses #334 and applies a stylistic consistency across all of OpenPNM, with a few exceptions, according to the PEP8 standard.
The exceptions are written in `tox.ini`, I decided to omit these files due to their incredible complexity and I didn't want to risk breaking any of these files, especially if we end up refactoring them


jgostick:
I will ask James to clean up his messy algorithms.  What should I tell him
to do?

maghighi:
I'll try to clean up GenericLinearTransport and GenericAlgorithm.

bspellio:
@RodericDay @jgostick This branch should be ready to go. It's a lot to look at, so try to just scan over it just to assure that I'm not changing any functionality.

bspellio:
@RodericDay @jgostick This branch should be ready to go. It's a lot to look at, so try to just scan over it just to assure that I'm not changing any functionality.

jgostick:
You've changed about 100000 lines of code, so looking it over seems
impossible.  I thought that was why we have tests, so we don't have to look
at this sort of thing? :-D

bspellio:
@jgostick of course, I was primarily referring to the `tox.ini` and `setup.py` because they have a few changes. The rest is pretty unimportant from an understanding perspective.

RodericDay:
I think rather than scan it over, you should just run it with whatever you're working with currently, and see if anything breaks.

If it does, that means more tests need to be written and things worked out. Otherwise, greenlight the merge.

PR #354, opened 2015-05-13
Testing framework

bspellio:
This PR implements our new unit and integration testing framework. The directory structure is as follows:
 * In `test/` there are two directories, `unit/` and `/integration`
 * In `unit/` the directory structure is exactly the same as the structure in the main OpenPNM directory with the exception that everything is a test class.
 * In `integration/` there is a series of scripts that will be converted to classes which will be used as examples/integration tests.

All of these tests can be run with  `py.test`

Additionally, after working with tox for a few days, I've decided it's not worth keeping it as a dependency. TravisCI doesn't seem to play well with it, and I find myself rarely using tox locally which was the intended purpose.


bspellio:
@maghighi @jgostick I'm going to want to merge this today once it's green. I've written several tests that you guys can use as examples.

Issue #356, opened 2015-05-14
labels() does not return correct values for throats


jgostick:
The problem is likely due to the comparison to empty lists

PR #359, opened 2015-05-14
Fixed 356 by changing how empty list arguments are detected.

jgostick:


bspellio:
@jgostick It's usually a good practice to reference the bug in the description of the PR, #356 , instead of just typing the corresponding number.

bspellio:
This looks good though, merge when green :+1:

Issue #325, opened 2015-04-01
numpy array comparisons


jgostick:
A while back I noticed that Numpy was throwing a warning about how comparisons (using == and !=) would be changing.  I think the time has come to clear this issue up, comb through the code and replace all possible offenders.  

The problem is that many arguments are [ ] by default, we then check the arguments to see if the default was overridden.  In the latest numpy pores != [ ] will now return a numpy arrary indicating were the condition is met, rather than a True or False check of the data type of pores.  

Two ways to avoid this:

    pores is not [ ]
    and
    sp.size(pores) > 0

The second is safer since we sometimes might convert the empty list to an ndarray with sp.array([ ],ndmin=1), the first option will be false, even though the data is actually an empty list.

jgostick:
Updates... @maghighi  and I noticed a few things worth considering:

    >>> len([])
    0
    >>> len([[]])
    1
    >>> sp.size([])
    0
    >>> sp.size([[]])
    0

So the behavior or len and sp.size are different for different shaped empty lists.  I think we need the behavior of sp.size.  

Also:

     >>> sp.array([]) != []
     sp.array([])
     >>> sp.array([1]) != []
     True

So why does this return an array sometimes and a bool others?

jgostick:
I think @TomTranter went through and did most of this already.  I have another 5 or so in core and GenericNetwork.  I think I've caught them all.  I used 'find in all files' to search for == [] and != [].  It took a while to figure out how the get Spyder to accept irregular expressions.

Anyway, I will create a PR.

jgostick:
There is ANOTHER issue with this...if the pores or throats arguments are boolean for some reason, then this size check will not return zero even if it is full of False's.  I've gone through the code and added boolean catch logic in several places, but I'm sure it's everywhere.

PR #360, opened 2015-05-14
Issue/325

jgostick:
This address issue #325 once and for all.  I think @TomTranter did a big batch of changes, but these catch the last few.

### DIFF
--- a/OpenPNM/Physics/__GenericPhysics__.py
+++ b/OpenPNM/Physics/__GenericPhysics__.py
@@ -63,7 +63,7 @@ class GenericPhysics(OpenPNM.Base.Core):
             self._phases.append(phase)  # Register phase with self
 
         if geometry is not None:
-            if (pores != []) or (throats != []):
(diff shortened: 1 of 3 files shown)


OpenPNM/Physics/__GenericPhysics__.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, network=None, phase=None, geometry=None,
             self._phases.append(phase)  # Register phase with self
 
         if geometry is not None:
-            if (pores != []) or (throats != []):
+            if (sp.size(pores) > 0) or (sp.size(throats) > 0):
```
bspellio: We don't actually need these parens.

jgostick: Let's keep'em, unless pep8 is being a little cry baby about them.  Bool
comparisons can behave very stragely without parens, in my experience.

bspellio: :+1:

bspellio:
:ship:

PR #370, opened 2015-05-21
Merge in voronoi voxel changes + pep8

TomTranter:


TomTranter:
Noticed a few things that need changing first and will try to improve coverage too

PR #372, opened 2015-06-01
Fix __str__ methods

bspellio:
During my PEP8ification of OpenPNM I broke a few string methods. This PR should that as well as doing an extra check when initializing a `ModelWraper`.
@maghighi @jgostick

### DIFF
--- a/OpenPNM/Base/__Controller__.py
+++ b/OpenPNM/Base/__Controller__.py
@@ -29,41 +29,45 @@ class Controller(dict):
         self.comments = 'Using OpenPNM ' + OpenPNM.__version__
 
     def __str__(self):
+        lines = []
+        horizontal_rule = 60 * '-'
         for net in self.networks():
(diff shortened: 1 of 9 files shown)


jgostick:
This doesn't seem right.  I haven't run it, but the changes to models dict seem a bit drastic, and what about Core, and the printable dict/lists?

jgostick:
Core prints fine, sorry about that...print printable lists don't

bspellio:
@jgostick The reason for the change is that `__str__` is supposed to return a string, not call print many times.

jgostick:
OK, understood.

jgostick:
Perhaps all of the __str__ methods should be converted to return a string?

bspellio:
They absolutely should be. I'll add that to this PR.

bspellio:
@jgostick good to merge this?

jgostick:
ok

PR #386, opened 2015-06-12
Fixes #385 by checking for the existence of the proposed name anywher…

### LINKED ISSUE #385: Creating multiple clones with same 'name'
When cloning simulations, its possible to specify the name of the new clone.  It causes a conflict if these are duplicated, so this should be checked.

jgostick:
Fixes #385 by checking for the existence of the proposed name anywhere within the simulation before creating the clones.


jgostick:
@maghighi can you check this and merge it?  It is bad form to merge your own PRs, so I'm trying to restrain myself :-)

maghighi:
I have already tested this one and it seems fine to me. I'll merge it now.

Issue #389, opened 2015-06-14
Things to delete?


jgostick:
There are somethings that never should have been added, and we 'may' be able to safely delete.

Utilities/shortcuts
Algorithms/Tortuosity
Algorithms/IP_for_imbibition

The algorithms can probably be moved to the new "resources" section on the website documentation?

jgostick:
I noticed that the code in the if __name__ == '__main__' section of files is not tested by py.test.  I don't think we have too much code in these sections, but I wonder if they're really needed, and should they be deleted?

jgostick:
return_results

Do we really need this?  Isn't it just better to have the user do it manually?  If there is any fancy business going on inside this method, then we can make methods that add it to the alg object dict instead...like throat fluxes in transport algorithms.  These are calculated by return_results, but should just be done by a method called calc_throat_fluxes, that adds a 'throat.flux' to the dict.  Then the user can pull out what they need.

jgostick:
I think we've trimmed it down a lot now, so I'll be closing this.

PR #392, opened 2015-06-16
Removing Legacy Code

bspellio:
This PR addresses #389 
We should not be keeping code in OpenPNM we don't plan on maintaining.
@jgostick @maghighi


bspellio:
So are we read to merge this?

PR #397, opened 2015-06-17
Fixed a bug in the late_pore_filling model

maghighi:
It was the same problem with the conduit conductance model. It was written with the old OpenPNM style.
In the new form, everything must be calculated for the phase, but only values of the physics should be returned.

### DIFF
--- a/OpenPNM/Physics/models/multiphase.py
+++ b/OpenPNM/Physics/models/multiphase.py
@@ -48,7 +48,7 @@ def conduit_conductance(physics, phase, network, throat_conductance,
     calculated.
 
     """
-    throats = phase.throats(physics.name)
+    throats = phase.Ts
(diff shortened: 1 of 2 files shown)


maghighi:
This PR fixes the same problem we had with conduit_conductance model. It was my bad I didn't catch it in the PR #394 . @jgostick, please check and merge it.

jgostick:
Perhaps you can write some unit tests for these models while you're digging around in there?  :-D  You can copy my examples for the phase models

maghighi:
I added tests for multphase models in the physics. It was a little tricky, especially checking the outputs of different conduit modes, but they seem to work.

jgostick:
Awesome, now I know that @maghighi knows how to write a test, so anoint him our new test guru :-D

PR #399, opened 2015-06-19
Fixed the test script

bspellio:
A small PR to ensure that we only need to maintain a single test command.
@jgostick

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - python setup.py install
 
 script:
-    - py.test -v --pep8 --cov=. --cov-config=.coveragerc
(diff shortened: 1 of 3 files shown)


jgostick:
Does this fix the problem we were having on Windows the other day?  If not, can you look into it?  Also what about the publish script, did you get that running too?

bspellio:
Yes I updated the publish script in a previous PR. As for the issue we were having, turns out windows doesn't natively have shebang support, the first line that reads:  `#!/usr/bin/env python3`. But you can execute the scripts on windows with [pylauncher](https://bitbucket.org/vinay.sajip/pylauncher), or simply with `python <script-name>`

PR #402, opened 2015-06-22
Feature/workspace

jgostick:
Moved controller functionality up to the package level


bspellio:
@jgostick Let's see if we can get this merged soon?

jgostick:
It passes all tests, and doesn't break @maghighi 's complex scripts, so it seems ok.

bspellio:
:+1:

PR #403, opened 2015-06-23
The addition of fixtures.

bspellio:
This PR adds a fixtures directory in `test/fixtures`. This directory should be used for any files that you would like to test against, e.g. and .pnm files. The path to this directory is made available through a global injected into each test called `FIXTURE_DIR`. Here's an example of how to load in a .pnm file.

```python
from os.path import join
from OpenPNM.Base import Controller
controller = Controller()

def test_load():
    file = controller.load(join(FIXTURE_DIR, 'test_net.pnm'))
```

Additionally, I've added a hook that runs after each test which creates a directory called `test/temp`. This is where you should save your files during your tests. Keep in mind, this file is completely deleted after each test and should be used to store valuable data. This directory is made available through the global `TEMP_DIR`.


bspellio:
@jgostick @maghighi

jgostick:
HA!  You got burned by PEP8.  I feel much better knowing that you too fall victim to your own rules.

jgostick:
OK, this looks great! I'll merge it, and we can give the group a rundown of how to use this functionality during tomorrow's telecon.

PR #408, opened 2015-06-25
Remove Shortcuts & Graphics

bspellio:
Do we have any objections for removing `Shortcuts` and `Graphics` from the code base? Shortcuts is not once imported or used though out the entire package, and Graphics depends on Python 2.7 which we are not supporting.

cc/ @jgostick


jgostick:
We should also remove the Graphic class, as this uses VTK, which is Python 2.7 only, and OpenPNM is 3+ only.  I never wanted to support heavy visualization anyway.

bspellio:
Great I'll add that to this PR

bspellio:
There are other parts of code that require that Graphics class and it is actually exposed to the user. Making it a breaking change. We should perhaps deprecate it for now?

jgostick:
The only parts of the code are the 'run script' that I know of.  Seriously, it's python 2.7, so it's already broken.

PR #414, opened 2015-07-02
Fixed minor bug in the solver

maghighi:
If the solver (not the algorithm object) was in a loop, it was using an
attribute from the first iteration and consequently the result would be
the same each time. Now it is fixed.


maghighi:
I guess this PR fixes the problem of solver in a loop. @jgostick, please check with your code, and if it is OK, merge it.

jgostick:
This fixes it. I can now clear the BC within a loop and reuse the same algorithm object.

PR #421, opened 2015-07-14
Awesome ip

jgostick:
This is an improved version of IP that tracks clusters of defending phase rather than invading phase.  The end result is that as the defender breaks up each cluster can be invaded in parallel.  The is useful for drying.


jgostick:
This is not really ready to go.  @maghighi has developed an improved version that will hopefully get merged in eventually.

Issue #427, opened 2015-08-21
Merge Pores Method


jgostick:
It would be handy if you could select a set of N pores, trim them and replace them with a single pore that is located at the geometric center of the original N pores, and connected to all the surviving neighbor pores.  This would an easy way (I think) to add multiscale-ness to the networks.

jgostick:
Added in PR #429

PR #429, opened 2015-08-21
Merge pores

jgostick:
This method was much easier to make than I thought, so I just went ahead and address #427.


jgostick:
Finally, the Travis build didn't break.

PR #433, opened 2015-09-04
Optimizing methods in GenericNetwork

jgostick:
Some of the methods in here haven't been looked at for a long time, so I'm seeing how many unnecessary and redundant lines I can get rid of.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 python:
 - '3.3'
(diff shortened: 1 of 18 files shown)


jgostick:
This PR actually worked...the find_neighbor_pores/throats methods are 10x faster now.  Individual calls are not noticable, but when called inside a for loop 1000 times the new code takes 0.8 seconds compared to 8 seconds.  If it passed the tests, we should merge it.

PR #438, opened 2015-10-29
Find connecting throats

jgostick:
This PR adds a method for finding the throat or throats that connects given pairs of pores

### DIFF
--- a/OpenPNM/Network/__GenericNetwork__.py
+++ b/OpenPNM/Network/__GenericNetwork__.py
@@ -300,17 +300,17 @@ class GenericNetwork(Core):
         Ts.reverse()
         return Ts
 
-    def find_connecting_throats(self, Ps, get_path=False):
+    def find_connecting_throats(self, pore_pairs, get_path=False):
         r"""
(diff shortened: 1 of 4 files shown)


jgostick:
Should include an option for assigning weighting for throat length or area values.  This could be used to apply saturation masks to the adjacency matrix.

jgostick:
Should we consider more general methods like 'find_neighbors', that accepts an argument whether to return pores or throats, or returns a named tuple with both in it?  This approach would reduce the number of methods, but make the methods more difficult.   We could also do 'find_path'

jgostick:
This method is working BUT the returned list of throats is not in the same order as the path.  They get sorted in the find_neighbor_throats method.  I'd like to fix this before merging.  Also, we still need to agree on the format of the return data.

jgostick:
So, the current approach is to place this find_path method in the Utilities.misc folder.  This way, the tool is there, but is not added irrevocably to the GenericNetwork class.  I suspect nobody will ever find it / use it where it is, but who knows.  We can always move it to the GenericNetwork class if it seem handy.

PR #441, opened 2015-11-03
Network importing

jgostick:
This is a basic but functional Import class.  
* The main principle was SIMPLICITY, so I thought having the data in a CSV format would be best.  
* This class accepts a file name and reads data in the CSV file into the Network object.  
* The headers of the CSV object need to be the property name with and underscore not a dot, like pore_volume and throat_length.  These become pore.volume and throat.length
* This file MUST have throat_conns and pore_coords, or else it will throw an exception.  These values are required by OpenPNM so this should be a surprize.
* each comma signifies a new column, so ND data (like pore.coords) must be entered with no commas such as 0.5 0.5 0.5, rather than 0.5, 0.5, 0.5.  This gets read as strings but the class converts these.
* the ability to specify ND data allows for things like throat.facets and other complex data, but this has not been implemented yet for anything other than pore.coords and throat.conns.


jgostick:
- Consider adding file format version and other info above the header
- Need to find a way to do info like [[1,2,3], [3,4,5], [6,7,8]]...this would be very hard to represent in the csv format

jgostick:
After the discussion during today's telecon, it might be worth while to look at NetworkX offers for output formats.  We should think about support one of them for importing...and possibly exporting too?  A quick glance suggests that their YAML format might be best...but there was't much describing the inner workings.  Perhaps Harlin could dig into it a bit?

PR #443, opened 2015-11-09
Rework docs

jgostick:
This PR moves the basic tutorial to the getting started position to help new users.  The former getting started is now called the inner workings.


jgostick:
I think I've made all the changes I want to for now.  I'll merge and regenerate the html tonight.

PR #444, opened 2015-11-09
Fix add boundaries

jgostick:
The current add_boundaries method (in Cubic) is a mess and offers no flexibility.  This PR introduces an add_boundaries2 that allows specific pores to have boundaries added to them.  It's not automatic, but it's fully flexible if you know what you're doing. 

We should discuss the API and if this is a good general format for other network types.


jgostick:
This PR now adds 2 new methods: 
1: add_boundary_pores, which is a fully flexible, but less convenient method for creating boundary pores
2: make_boundaries_period, which takes 2 list of boundary pores and links them into nearest pairs

PR #452, opened 2015-11-24
Updated transformation.py

jgostick:
…e.  The copyright is 2015, not 2014, so maybe the broken tests work now?


jgostick:
This is a mess.  I have ask @TomTranter to see if the transform3d package will do the same job, without having to maintain (and test) the code in the transforms.py file.

TomTranter:
I only use 5 functions in the transformations file. I've had a look at transformations3d and it pretty much has the transformations.py file copied and pasted into the package as _gohlketransforms. It has some pep8 changes which is what I guess is causing you problems. I think we can use it. I'll make some changes in this branch

TomTranter:
@jgostick the travis build failed as it couldn't find transforms3d on the server. I included it in the setup.py file and thought that was all that was needed...

jgostick:
I there it needs to be in the .travis.yaml file too.  I'll add it.

jgostick:
OK, all good.  I added a line to the travis.yml file to do a pip install of transforms3d.  According to the travis website, this should not have worked but it did.  

Also, not sure why the coverage went down so much?!  Did you remove the transforms.py file?

TomTranter:
Yes which was well covered so probably brought down the average


---- jgostick wrote ----


OK, all good. I added a line to the travis.yml file to do a pip install of transforms3d. According to the travis website, this should not have worked but it did.

Also, not sure why the coverage went down so much?! Did you remove the transforms.py file?

-

PR #457, opened 2015-11-26
Add units tests

jgostick:


jgostick:
A few of the save/load tests that I added don't save to the TEMP_DIR...I will fix that next push.  Sadly, these tests don't improve the damn coverage!

Issue #459, opened 2015-12-01
Access Limited OP bug


TomTranter:
I think there's a bug where if access limited is set to False non-invaded clusters are included in the invaded_clusters. This would be a one line change to put the following line outside the if statement in _do_one_iteration()
inv_clusters = inv_clusters[inv_clusters >= 0]
Whilst I've been looking at this though it occurred to me that we should develop it to do bond and site percolation as it currently just does bond or we should rename it drainage.

jgostick:
The new version of find_connected_clusters does do both.  You're right though, OP should be called bond_percolation.  It's too late to change that really, so we could add an argument to the algorithm that indicates which type of percolation to do.  

Also, I'm not surprized that there is a bug in the access limited = False code.  It was really only tested for access limited cases.  Open a PR, make the change, add some unit tests.

TomTranter:
@jgostick the capillary pressure model calculates the entry pressure for throats. For site percolation it should calculate it for pores. This highlights a general problem with our models framework where some models are pore or throat specific and some are more general and could apply to both. With the geometry models there is usually a clear distinction with the model name containing the word pore or throat. I think for clarity and ease of use it would be good to standardize the models along these lines. For example throat only models would start with the word throat, pore only models would start with the word pore and general models would start with the word element or entity and have a mandatory string argument  being either 'pore' or throat' and maybe also 'both' to tell the model which one to calculate.

jgostick:
I agree on all points.  We can discuss more during the telecon.   For now I will create a PR with two new algorithms, one called primary_drainage and one called secondary_drainage that starts with some pores pre-filled.  Both will assume access limited and trapping will be an option.

TomTranter:
OK  - I've been working on OP in my repo to fix the original problem I found and I've already made some changes to allow it to do bond or site percolation. If you think a smaller simpler algorithm to do each sort of percolation is the way to go with fewer arguments then that's fine. I can merge my work into a branch on PMEAL or discard it

TomTranter:
Fixed as part of #466

PR #466, opened 2015-12-08
Drainage algs

jgostick:
This is the beginning of a set of changes proposed to the drainage algorithms as pointed out in issue #459.  

Eventually, this PR will create a more simplified drainage algorithm (including trapping and residual phase), and also an imbibition algorithm that uses pore sizes instead of throat sizes.

### DIFF
--- a/OpenPNM/Algorithms/__Drainage__.py
+++ b/OpenPNM/Algorithms/__Drainage__.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+"""
+===============================================================================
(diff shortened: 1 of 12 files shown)


jgostick:
This is now DONE.  The test coverage is nearly 100%.  I think the code works as well as I hoped.

jhinebau:
Hey Jeff. I don't know if you added trapping to drainage yet, but if not, I accidentally asked 15 students to develop trapping algorithms a few years ago (major nightmare). It resulted in some very interesting techniques. They gave me their code (Matlab), so I can pass that along to you if you are interested.
Cheers,
James

PR #483, opened 2016-01-12
General io network

jgostick:
This PR Merges two existing PRs: #481 and #441.  The intention is to write a variety of IO methods (Utilities.IO) that save and load data from certain file formats, AND combine these IO methods with a Network generation class that can pull data in from files to populate a Network object.


jgostick:
This is almost ready to merge.  There are few things to discuss at the next telecon:
1.  Do we all agree with the formatting in each file type (it's a bit different in each, is this bad?)
2.  The use of pipe delimiter for phase data...should the load methods parse this data into phase objects?
3.  Amalgamate data is heading for the dust bin due to the new props(mode=['deep']) method

PR #492, opened 2016-01-30
Statoil io

jgostick:
This PR adds the abilitiy to import the so called 'Statoil' file format into OpenPNM.  These files are used by the Blunt group's network extraction algorithm, and presumably some other code at Statoil.  Unit tests are run on two sets of data from the [ICL website](https://www.imperial.ac.uk/engineering/departments/earth-science/research/research-groups/perm/research/pore-scale-modelling/micro-ct-images-and-networks/).


jgostick:
In it's current form:
* I have removed the Import class
* Added import_data and export_data to the controller class (and also help methods at the package level)
* These methods always return a new network object...the user must manually move data around as needed
* The actual IO classes optionally accept a Network, and if received any existing data is NOT overwritten.  This allows users to have more control over the processes if they're willing/able to dig deeper into the code
* The YAML class was renamed to NetworkX since that is really what the class if for
* The NetworkX and Statoil IO methods ARE not accessible from the controller, only from the Utilities.IO
* Keeping the controller methods only for general file types made the most sense to me.  
* I also removed the 'example_net.mat' file from the Network module, moved it to fixtures, and changed the unit tests accordingly.

Still needed:
* An example is being added to the 'improved_documentation' PR
* We should also add some proper definitions about file formats for each of our supported file types.  This isn't suitable as an example, so should be added to the user guide somewhere?  Again, I think i'll roll this into the 'improved_documentation' PR.

Bottom Line: This PR is ready to be merged, and we should aim to tag the master branch with V1.4 during the next telecon.

Issue #496, opened 2016-02-03
Rename Controller to Workspace


jgostick:
Calling it the Controller is very confusing and Workspace is exactly what it does. We can do this with backwards compatibility no problem at all.

jgostick:
This is a simple change and mostly involves updating the documentation to reflect the improved terminology.  I think it should be a V1.5 thing, so that it's ready for Interpore.

PR #497, opened 2016-02-03
Workspace

jgostick:
This PR is to address issue: #496 to rename the Controller class to Workspace.  This name change will make it much more obvious what the class is for.  It is also setting up for the day when there is a Simulation class that acts as a container for sets of Core objects.


jgostick:
The fact this works is very positive considering that we have a test that loads a PNM file from V1.2.  If that test doesn't break then this is pretty solid.

jgostick:
Unfortunately, this PR is very old and has MANY conflicts with the current code.  I"m going to scrap it and start over.

PR #504, opened 2016-02-10
Improved pore diameter models

jgostick:
This PR adds numerous pore and throat diameter models that are explicitly named for the distributions they use (normal and weibull).  This is to make it more clear to users what they are getting.


jgostick:
Write unit tests for other models to be thorough.  Also, there are still a few fixes to docstrings.

Issue #506, opened 2016-02-11
An Empty Network class


jgostick:
It might be handy to have a Network class called Empty that accepts Np and Nt arguments.  It would then return an object that is a bit more functional than GenericNetwork since you could use the standard dict assignments: pn['pore.blah'] = 4 and pn['throat.bleep'] = sp.ones(pn.Nt)

PR #509, opened 2016-02-11
Empty network class

jgostick:
Just a small PR that adds and Empty class to the Network submodule. The point is to create a Network object that be used to incrementally add data using the standard dictionary syntax and benefit from the Core data management tools.  The current option is to use a GenericNetwork, then use the ``dict.update`` method to bypass the Core data management rules.  

This fixed issue #506.


jgostick:
Odd that Travis didn't build this PR...

jgostick:
Before merging this, we should consider just adding Np and Nt as optional arguments to GenericNetwork

Issue #467, opened 2015-12-09
Move set_boundary_conditions to GenericLinearTransport?


jgostick:
This method is currently in GenericAlgorithm but it's highly geared towards transport algorithms.  I think I originally moved it up to this GenericAlgorithm level, but now I'm changing my mind.  The reason is that I'm writing a Drainage algorithm and could use this method but it would be completely overloaded/overwritten.  I think the method should be an non-implemented method at the GenericAlgorithm level.

PR #515, opened 2016-02-16
modified solver

maghighi:
This pull request addresses the issue #467. Also, the unit and integration tests have been modified according to the new changes.


jgostick:
Why didn't Travis run the tests?

maghighi:
I'm not sure. I ran all the tests locally and then made the pull request. But Travis didn't run the tests and I couldn't figure out why.

jgostick:
This happened to me on the Empty Network PR, because I'd named two tests the same name (I think).  I'll check it out.

jgostick:
Or perhaps it thinks nothing has changed since you just cut and pasted code around?

maghighi:
I moved the code from generic_algorithm, but the code in the solver is totally different, because I had to remove the component from everywhere. Even if it was exactly the same, Travis should run the tests, because the fact that something runs properly in a generic class, does not guarantee that it can be run in its subclass.

PR #537, opened 2016-03-09
Adding a mode to clear that allows deletion of all empty labels.

jgostick:
This is a tiny tweak to ```clear``` to allow removal of all empty (defunct) labels using the ```obj.clear('empty_labels')```


jgostick:
not worth adding the cruft to the  Core class

PR #540, opened 2016-03-16
Fixing setup.py

jgostick:
Changed setup.py to stop circular logic of importing package before i…nstalling it.  Changed how the version number is retrieved.


jgostick:
I'm going to go ahead and merge this, tag it 1.4.5 and push it to PyPI asap since it's costing me sleep that the installer is broken.

PR #543, opened 2016-03-23
Move topology tools

jgostick:
Moved all the tools that were in Utilities.__topology__ to Network.tools.  
- These are now all 'functions' and not inside a class so there is no need to instantiate a 'topo' object to access them.  
- Helper methods have been placed into the old __topology__ file to support legacy calls
- These helper methods are all now static methods so you don't have to instantiate at ``topo`` object, but you can if you want.  
- This is listed as a goal for V2, but I think it could be included in V1.5, and then the helper methods removed from v2.


jgostick:
This move was agree to during the telecon today, so I'll merge it now

PR #544, opened 2016-03-23
Tiny bug fixes

TomTranter:


jgostick:
Looks good, you basically did a search for print and replaced with logger.  Good idea.

PR #550, opened 2016-04-06
Adding physical_constants to GenericPhase

jgostick:
A first pass at adding a physical_constants attribute, and populating it with the obvious properties.


jgostick:
I realized after the first commit that this is a perfect example of a sandbox feature, SO I created the sandbox file and moved the PhysicalConstants class there.  You can use it by doing:

    from OpenPNM.Utilities.sandbox import PhysicalConstants
    air.physical_constants = sandbox.PhysicalConstants()

jgostick:
No one could think of a compelling reason to do this, so let's close it

PR #552, opened 2016-04-07
Changed from pickle to dill

jgostick:
Using Dill for saving and loading files should be MUCH better as it has the ability to pickle custom functions, not just functions in the openpnm package.


jgostick:
Dill is a drop-in replacement for pickle with extra functionality.  All files saved from now on will be able to save and reload custom models, not just ones from the openpnm library.  It also means that we can change and remove models without breaking old files anymore (sorry @TomTranter)

jgostick:
Before merging this branch, I want to make sure that dill does actually store the model objects in the file.  The unit test I wrote might not actually check this properly.

jgostick:
I have just confirmed with certainty that dill works where pickle does not.  I created a run script with a pore-scale model definition in it.  It saved with pickle, but threw an error when attempting to reload.  When I imported dill as _pickle it saved and loaded just fine.  Therefore, this PR is important.

Issue #555, opened 2016-04-13
check_network_health not catching all duplicates


stadelmanma:
When reading the test network from iMorph, the function doesn't seem to catch all of the duplicate throats in the first run. I don't understand the theory used to track down duplicate throats well enough to tell why it's missing so many throats. The attached spreadsheet has 3 sheets the first is an overview of the node file, the second is the duplicate groups catch by check_network_health and the third sheet is the duplicate groups I've confirmed using excel and my removal function. 

Line 1067 in IO.py is where I set duplicate throats to 'None' so they get filtered out, if you comment out that line you can reproduce the error using the upload found in test/iMorphTest in my repo.

[imorph_net_dup_throats.xlsx](https://github.com/PMEAL/OpenPNM/files/217452/imorph_net_dup_throats.xlsx)

jgostick:
I've looked into the tests, and we do have some basic unit tests for the ```check_network_health``` method, but they only do a few throats, not a whole pile.  I've been playing with the code from your repo and I am seeing the issue you speak of.  I'll keep checking.

xu-kai-xu:
Hi, @jgostick , I just construct a locally-connected pnm, and run `check_data_health`. The checking result is easy to understand except `duplicate_throats`. I am not sure what duplicate means. Does it mean two `pn.conns` arrays are repetitive, or something else.
![image](https://user-images.githubusercontent.com/44153643/100222372-f0e17d80-2f54-11eb-970c-26d39d80082a.png)
The pnms includes 785 pores, 785 throats, and 225 local clusters. The `duplicate throats` array includes also 225 throats. That confuses me.
Thanks for your help.

jgostick:
Yes, duplicate throats means that you have [i, j] in the adjacency matrix twice.  Bidirectional throats means you have [i, j] and [j, i] in there.  If you do ``op.topotools.trim(network=pn, throats=h['duplicate_throats']`` it should clear things up.

jgostick:
Looking at your picture I see lots of 3-pore + 2-throat, or 2-pore + 1-throat little clusters, so the number of throats in this case should be < the number of pores.

xu-kai-xu:
Yes, I made a mistake. There are some duplicate and bidirectional throats. After the trimming operation, the throat number decreases to 560.

PR #556, opened 2016-04-15
Changed the way duplicate throats are found.  

jgostick:
Fixed issue #555.  It was finding correct throats, but apparently not all of them.  This fix works, but may not be fast as it converts to sets inside a for loop.  Optimize later if needed.


jgostick:
I just confirmed that this works exactly the same as the old version (i.e. the returned list contains not duplicate sets), so I'm going to merge.

PR #560, opened 2016-04-22
Trying to get codecov working again.

jgostick:
I will be doing a lot of trial and error to see why codecov hasn't been working.


jgostick:
Yay, it works!  I think that the 'clean' command was destroying the coverage info before it was send to codecov.  I also enable coverage checks even if the build fails.  There was a problem before that coverage would not kick in if the first build failed, so this may fix it.

PR #561, opened 2016-04-24
Remove cruft

jgostick:
Removing several unused lines/blocks of code, and adding some unit tests


jgostick:
w00t!  1.5% increase in coverage.

PR #568, opened 2016-05-06
Imorph upload

stadelmanma:
The upload could still use a few tweaks but it is fully functional. I removed all the network cleaning I used to do and just replaced it with a call to `check_network_health` to notify user if there are issues. 

Boundary_cell_face pores are turned into zero volume pores with props transferred to the throats connecting them to the network. They get labels stating which face they pertain to as if `add_boundaries` was called on the network. 

Last changes are adding in micron/voxel to meter unit conversions and a little error handling on the inputs.


jgostick:
Awsome, thanks for the PR!  I hate to say this, but the build failed due to PEP8 issues.  You can see where the errors are on [Travis](https://travis-ci.org/PMEAL/OpenPNM/jobs/128379975).  I get nailed by pep8 errors all the time.

jgostick:
Nice work, but merging this will decrease test coverage by 1.2% (boo!).  You need to write some units tests.  Unfortunately even though you've made a PR I can't edit your repo, otherwise I'd get them started for you.  Instead, you'll have to do it, but here's the basic idea:  In the file "OpenPNM/test/unit/Utilities/IOTest.py" you add a function call ```test_imorph_load```.  In that function you setup a very basic simulation where you read in an actual iMorph file.  Save the fail in the "OpenPNM/test/fixtures" directory (don't forget to add this to the commit!).  Use the FIXTURE_DIR global variable to open the file, then test that the return network has all the specs you expects (i.e. Np, Nt, etc).  Use the other functions as a template obviously.  Sorry to do this to you, but it's the unofficial rules: must write tests!

stadelmanma:
The tests were easy to write, I just wasn't sure if that was something that was handled on your end or mine. I think it is ready, if there are any other changes you want to make after reviewing it let me know and I'll get them in.

jgostick:
Nice work.  It's best to you do the unit tests since you know the code best.  You got 95% coverage so that brings up our average ;-)  I'm at interpore this week, but I"ll try review and merge it at some point.

jgostick:
@stadelmanma I have a suggestion for the PR: i think the argument for the load method should be ```path``` that points to the folder containing the nodes and graph files.  As it is now it requires two arguments that are long and tedious to type, and I 'think' they are always the same, right?

stadelmanma:
In theory, if the user doesn't change anything from the export iMorph creates all that would be required is a path to the folder location since I am pretty sure the file names are always the same.  I wasn't sure exactly where to draw the line between convenience and flexibility, so I tended towards flexible. 

 What I have been typically doing is: 
`path = \path\to\folder\`
`io.iMorph.load(node_file=path+'node_fname', graph_file=path+'graph_fname')`

 So I suppose we have three options: 

`A. io.iMorph.load(path='path/to/folder')` 
`B. io.iMorph.load(path='path/to/folder', node_file='node_fname' graph_file='graph_fname')`
`C. io.iMorph.load(node_file='path1/node_fname' graph_file='path2/graph_fname') #current`

A. is the most convenient but assumes the user and iMorph always use the same file names
B. is a middle ground and forces the files to be in the same directory
C. lets you get files from anywhere, any name, I like the flexibility since it allows a user to vary the input files as desired (assuming the nodes are the same, maybe scaling pore/throat sizes?). 

I am fine with any of the above options, so what ever you think would be best for the code base as a whole.

jgostick:
How about D: ```io.iMorph.load(path, node_file="throats_cellsThroatsGraph_Nodes.txt", graph_file="throats_cellsThroatsGraph.txt")```

Where must be set by user, and the other two default to the expected file names?

stadelmanma:
I went ahead and added your version of load arguments since I thought that was a good compromise. 

Also your comment about `Empty` and `GenericNetwork` networks got me thinking about the upload. Right now I always recreate an empty network regardless if one was supplied to prevent errors. What are odds of that causing a major error if someone for example is using the name of the network? I tried a few things but there wasn't a _quick_ fix since we need the actual network methods where as the other uploads only need to store the data in a dictionary.

jgostick:
I agree that someone is unlikely to pass a network as an argument, but you never know.  I think the logic behind that was to allow people to import some information form some places and then additional information from others...like network connectivity from NetworkX then geometry from CSV or Mat-file.  

This is looking great now.  We should merge it and tag V1.5 as soon as a few other PRs are passing.

PR #569, opened 2016-05-08
More doc refinements

jgostick:
I merged this last week to help ReadTheDocs and Codecov get the correct info from the Master branch, but I need to keep working on the docs so I'm starting a new PR.


jgostick:
The docs on 'readthedocs' are currently incomplete.  I think we need to finish them before we can call version 1.6 official.  I"m tagging this as v1.6 and will work on it in a few weeks.

PR #571, opened 2016-05-13
Op site percolation

TomTranter:
These changes allow OrdinaryPercolation to run in "bond" or "site" mode with a new "percolation_type" parameter. Pore entry pressure is also added as a parameter and some capillary pressure models have been changed to be more generic allowing the calculation for pores or throats.


TomTranter:
@jgostick, not really sure why tests are failing on python 3.5 on the last commit but were fine for the one before when the tests that are failing are nothing to do with the code that changed. Maybe a random variable on the server found a but. Can you re-run the travis tests for a commit?

jgostick:
yes, i can rerun when logged in...willdo now

jgostick:
@TomTranter Is this good to merge?  There are a lot of changes!

TomTranter:
@jgostick it will break scripts that use capillary pressure and define a throat_diameter variable instead of taking the default. I renamed the variable diameter as the models now work for pores and throats. We could add some backwards compatibility by accepting a keyword argument. Other than that.. if all the tests are passing...

TomTranter:
Quite a few of the commits you see were later reversed so it's best to just look at the code changes

TomTranter:
@jgostick I mis-remembered. It doesn't break the scripts - OP defaults to bond percolation if a pore.capillary_pressure data is not found on the phase. And the capillary pressure models would also default to diameter='throat' so it handles legacy types of call to the model and algorithm. It's not entirely clear to the user what has been run although there are a few warning messages. One new thing I've just added is a small function to handle boundary pores. If creating one physics for the network including boundary pores then the zero diameters are handled by setting them to equal the largest pore in the network.

TomTranter:
@jgostick I think we should merge this. I also want to upload a bunch of new changes to run the hysteresis code so it would be good to get this one out the way instead of branching off a branch

TomTranter:
@jgostick what's with all the errors? They seem to be mostly to do with the workspace...

TomTranter:
@jgostick it looks like some functions are trying to access geometries from the network and they are no longer there. Did you merge in a version 2 change?

jgostick:
I don't know what the hell happened...it actaully passes all tests on my local machine.  I 'think' it has something to do with the latest version of scipy which apparently had a larger than normal number of breaking changes.

TomTranter:
@jgostick tests are all passing now on this branch. I commented out that little change to clear in workspace as it was the culprit but it would be worth creating a new branch to work out why it caused so many tests to fail and fix the memory leak.

Issue #580, opened 2016-06-05
set_boundary_conditions doesn't accept boolean mask of pores


jgostick:
This could/should be addressed using teh _parse_locations method that was added to core

PR #582, opened 2016-06-06
fixed_boolean_mask_BC

maghighi:
This pull_request addresses the issue  #580. I have also added some more
unit tests to cover the issue.


jgostick:
@maghighi you'd better write some more tests so this passes :-D  Also, while you're at it could you beef up the doc string for ````set_boundary_conditions``` and ```set_source_term```?  I'm thinking examples at the bottom, and maybe a bit more explanatory text about each option?

PR #583, opened 2016-06-07
Adding the bare outline of an Interface class. 

jgostick:
It is currently stored in Utilities.sandbox, since this is the perfect time to add this.


jgostick:
This discussion must wait for V2, so closing this

Issue #527, opened 2016-02-29
Add a 'find_surface_pores' function


jgostick:
In issue #484 is was pointed out that adding boundaries after stitching can be a pain.  That issue was closed since the recently added 'add_boundary_pores' method address it, but there is still a need to have a function that can label all surface pores.  There is a function like this already in the Utilities.topology file, but perhaps it should be generalized and cleaned up?

jgostick:
OR, how about this:
Put pore centers in an image, do a watershed transform and find which pores 'drain' to the edges of the image...these are all the surface pores.  You could even find which face they belong to by seeing which face each drainage basin flows to.

PR #585, opened 2016-06-09
Find surface pores

jgostick:
Added a function to Network.tools that finds pores on the surface of any network and labels them as 'pore.surface', to address issue #527.  This method uses a watershed transport to segment the space around each pore center, then labels all pores whose watershed basin touches the edge of the image as a surface pore.  Because it uses image analysis tricks, it can be slow in some cases: particularly subdivided networks.

### DIFF
--- a/OpenPNM/Network/tools.py
+++ b/OpenPNM/Network/tools.py
@@ -800,20 +800,3 @@ def find_surface_pores(network, im_max=1e7):
     surface = _sp.unique(labels[inds])-1
     network['pore.surface'] = False
     network['pore.surface'][surface] = True
-
-
-
-
-
-
-
-
-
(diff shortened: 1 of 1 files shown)


jgostick:
Here is a shot of it in action:

jgostick:
@ma-sadeghi and @maghighi, can you check this PR on some of your fancy subdivided and odd shaped networks to make sure it catches the surface pores that you think it should?  Let me know if the documentation makes sense and it provides the functionality you need.  

Also, check it for speed on larger networks.  At the moment it performs the Delaunay tessellation multiple times.  I tried doing it once and reusing it, but for some very odd reason it would literally stop my python interpreter!

Issue #574, opened 2016-05-26
Add Voronoi Network class?  


jgostick:
The Delaunay Network is nice, but numerous papers have used the Voronoi network as the pore network.  @jhinebau has published something like this.

jgostick:
Better yet, we've decided to make a DelaunayVoronoiDual that create both plus connections between them.  This class will be used to make just a Voronoi by deleting the Delaunay pores, and vice versa.  A PR has been created.

PR #593, opened 2016-06-14
Add voronoi network

jgostick:
This creates a set of base points and find the voronoi tessellation around them, then converts the voronoi tessellation into the pore network.  It is the compliment to the Delaunay network.  This addresses issue #574.

### DIFF
--- a/OpenPNM/Geometry/models/pore_diameter.py
+++ b/OpenPNM/Geometry/models/pore_diameter.py
@@ -57,14 +57,35 @@ def random(geometry, seed=None, num_range=[0, 1], **kwargs):
 
 def largest_sphere(geometry, network, **kwargs):
     r"""
-    """
+    Find the maximum diameter pore that can be place in each location that
+    does not overlap with any neighbors.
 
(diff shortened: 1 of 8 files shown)


jgostick:
This isn't quite ready to merge yet...we need to discuss the various options/behavior we want.  Also, sometimes the connections don't quite look right.

jgostick:
New and improved now with awesome [screenshot!](http://i.imgur.com/9xhaZxG.png)

jgostick:
This works fairly nicely now, with two glitches:
1. The voronoi cells near the 0 axes tend to have vertices just less than 0.  The ones on the other faces are exactly on the face...this should be fixed, but I'm not sure what's causing it.  At the moment I just keep facets that are within -0.1, but it's a bit arbitrary.
2.  How to robustly add boundary pores to this network. Also, I"m not labeling the surfaces or faces yet, but this reason.

jgostick:
This PR is now defunct with the new DelaunayVoronoiDual class.

PR #594, opened 2016-06-16
Largest sphere model

jgostick:
Totally awesome pore scale model to find the largest sphere you can put in each pore without over lapping any neighboring pores.


jgostick:
This isn't quite ready yet.  I need to adjust it so that it works on a subset Geometry, while for now it doesn't treat the pores on neighboring Geometries correctly.

PR #595, opened 2016-06-16
Fixing TravisCI Python3.3 build error

jgostick:
Seeing what happens when atlas and fortran compilers are not installed...anaconda should be coming with a binary of numpy so these aren't needed.


jgostick:
Not only did this fix the broken build, but not installing the atlas libs saved a few minutes off the build time!

PR #596, opened 2016-06-16
Allowing for hidden or private dictionary entries

jgostick:
Adjusted the __str__ function on Core to not show any dict keys starting with 'pore._' or throat._'.  Also adjust the PrintableList to do the same.  This way print(pn), print(pn.props()), and print(pn.labels()) will all exclude the hidden or private arrays.

Actually implementing this throughout the code will be a pain, but at least the framework is ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,6 @@ script:
 
 after_success:
   - codecov
+
+notifications:
+  email: false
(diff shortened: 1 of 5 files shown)


jgostick:
On second though, let's not change the way objects are named, to keep everything constant, and just start using this feature in the algorithms.

PR #598, opened 2016-06-22
Rename_save_to_save_workspace

jgostick:
Renamed save and load to save_workspace and load_workspace to be more descriptive.  Change is backward compatible, except that the helper methods at the package level were also updated, but not made backward compatible.


jgostick:
I'm going to merge this even though CodeCov is complaining.  There are no tests for the old save and load which are now deprecated.

Issue #597, opened 2016-06-22
Agenda for June 22, 2016 Telecon


jgostick:
* Discuss & Merge some PRs ( #585 , #596 , #594 )
* What's up with OP Site Percolation ( #571 )

TomTranter:
Hi,
I'm not able to join the meeting today. The problem with OP Site Percolation is a lack of tests covering the changed lines. I've been a bit busy to do this but will see to it next week in time for the telcon so we can discuss the changes

Issue #553, opened 2016-04-09
Plot pore coords in 3d


jgostick:
It might be handy if the GenericNetwork class had a method for plotting the pores in 3D, so you don't have to open paraview all the time.  I tried using matplotlib's 3d plotting functions ages ago and they did not work.  I'm sure that has changed...in fact I think @TomTranter has some such function in the Delaunay Voronoi example.

PR #600, opened 2016-06-23
add_quick_network_visualization_plot

jgostick:
Got the 3d scatter and 3d line plots working quickly using some insane vectorization.


jgostick:
This address issue/feature request #553

jgostick:
@ma-sadeghi Can you try running this on your fancy workstation to see if you can do a big network?  It seems to generate the plot very quickly, but the moving and rotating the points is slow, which depends on the video card of course.

ma-sadeghi:
It was fast in terms of runtime (took about a second to plot a 15 by 15 by 15 network). However, moving and rotating the points is still so slow (even for 15 by 15 by 15).

jgostick:
That's odd, it's better that than for me on my laptop I think.  I did 25^3 and it plots in just a second or so, and rotation is laggy but only by 1-2 seconds.

ma-sadeghi:
Does your function return a handle to the figure object?

Get Outlook for iOS




On Thu, Jun 23, 2016 at 3:13 PM -0400, "jgostick" <notifications@github.com> wrote:





That's odd, it's better that than for me on my laptop I think.  I did 25^3 and it plots in just a second or so, and rotation is laggy but only by 1-2 seconds.

---

jgostick:
Yes it does.  My spyder IDE is set to automatically show all figures though so I don't have to do show() or anything. Why do you ask?

ma-sadeghi:
Cuz it didn't work when I tried using a text editor (Sublime and Atom), even after forcing it to show.

Get Outlook for iOS




On Thu, Jun 23, 2016 at 3:24 PM -0400, "jgostick" <notifications@github.com> wrote:





Yes it does.  My spyder IDE is set to automatically show all figures though so I don't have to do show() or anything. Why do you ask?

---

jgostick:
Hmmm, you may have to call plt.ion() to make it interactive?

jgostick:
@ma-sadeghi I totally changed the logic, so it should be much faster now.  It also accepts a list of specific throats to plot if you want to zoom into a sub-region.  Can you try it out on the workstation again?  It's decently fast on my laptop with 35*25*15 pores.

ma-sadeghi:
Nice try. It's now way faster. Still laggy on a network of 100,000 size, but OK. Anyway, I figured the problem for the plot not showing up. This should do the job:
```python
import pylab
pylab.show()
```
I think it's a good idea to inform the user of this.

jgostick:
I guess the final question is where to put this method.  Should it be moved to Utilities.Plots along with the other plotting methods (```geom.plot_histograms```)?  Should it be attached to the generic network like ```network.plot_topology()```?  Maybe the ```plot_histograms``` method should be moved to Geometry.tools which doesn't exist yet?  

I'll add it to the next telecon agenda #597.

PR #604, opened 2016-07-07
Delaunay voronoi dual

jgostick:
A class that creates duel Delaunay and Voronoi networks, and connects them to each other to allow interaction.


jgostick:
Pretty pictures of the voronoi pores and throats only on the surface:

jgostick:
@TomTranter First of all, apologies for duplicating so much of your code, but I was having fun and learning a lot about the Delaunay and Voronoi functions in Scipy.  As discussed, the eventual aim of this class is to also make a Voronoi and a Delaunay class that uses this, then trims the un-needed pores/throats.  Can you remind me what sort of 'special' information you were storing on the Delaunay that was then being used by the Voronoi geometry class?

jgostick:
I decided not to change the names of the Voronoi and Delaunay networks, or allow user specified names.  It involved a lot of changes in the code and I didn't want to make it more complicated than it already it.  Also, it's quite clear what Delaunay and Voronoi mean so there is no need for users to change it.

PR #607, opened 2016-07-18
Viscous drainage

stadelmanma:
I still need to write unit and integration tests for this as well as more general clean-up but I'm going to go ahead and PR so it's more visible to anyone else interested.


stadelmanma:
@jgostick I added a basic integration test more as a template for you to play with if your curious and haven't had a chance yet. I would still like to do a good bit more tests with the new algorithm to make sure I have worked out as many bugs as possible that would cause a simulation to stall.

 The main one I remember is when there are multiple zero volume boundary pores attached to a single real pore, not an issue for your network generation but iMorph in particular adds multiple BPs to real pores that are cut off by the edges of the ROI. 

Other things that need hammered out are runtime logging, file generation, allowing users to add their own exit & logging criteria with relative ease, etc. It will be a little while before I am able to return to this full time, probably mid-august.

jgostick:
mid august is probably when I'll be able to return to coding myself.
Thanks for all this!

On Tue, Jul 26, 2016 at 9:13 AM, Matthew Stadelman <notifications@github.com

jgostick:
@stadelmanma I have not forgotten about this, but I am over my head in grant writing the moment.  Will review it in a few more weeks.

stadelmanma:
@jgostick  No worries, I have been pulled back into fracture flow modeling since that last comment I made, and probably will be doing that until the end of the year. So I haven't had the chance to do a final debugging and checkout all the corner cases that could cause the algorithm to stall out. 

When you have the time to return to OpenPNM actively let me know and I'll start running some cases on the side when I can to begin the testing process.

jgostick:
Hey @stadelmanma, I'm reviewing this right now, and will hopefully be done today.  I want to include it in V1.6 which we'll release before the year's end.  Github says that you need to update your branch, so go ahead and do that.

stadelmanma:
hi @jgostick that is exciting. I would recommend running the code on a few small networks both artificially generated and some real ones to see how it performs. If I remember right it had a tendency to move very slowly on some networks because the code would stall out essentially fighting over a throat, especially on regular cubic networks. To an extent that is just a limitation of the IMPES simulation method. 

Alot of the stuff it prints out is debugging messages since I was tracking throat and pore filling to make sure things ran correctly and probably can be removed. As well as using traditional Python logging instead of the temporary function I made. 

I am terribly bogged down right now but I will give the code some test runs and clean out any oddities up before the end of next week when things will hopefully have slowed down a tad.

jgostick:
OK, so I think the way to do this if for me to fork your repo, make changes
and ask for you for a PR.  Then you merge any changes then ask OpenPNM for
a PR.  Convoluted, but unavoidable.  I'll start by reading the paper
carefully, running your code, and playing around a bit.

jgostick:
@stadelmanma We're going to merge your PR into OpenPNM.  It's great code and the underlying algorithm is based on solid research.  My main concern has been the responsibility of maintaining some one else's massive code, but we figure if it ever stops working we must've done something bad to the code.  Anyway, we need you to re-merge openpnm into your branch, then update the PR, then I'll merge it.

TomTranter:
Hi @stadelmanma, I've taken your integration test as an example and I'm changing some of the parameters to try and better understand the algorithm and also to try and produce different results. 

I think we definitely need an example with a realistic network, the integration network has throats which are very long compared to the pore sizes and also is quite small (understandably for a test). I've taken the lattice spacing down a factor of 10 so that throat and pore sizes are comparable.

If I change the flow rate to be very small so that capillary forces dominate it runs forever. It seems like the logic on the time stepping could be improved as the algorithm should reduce to ordinary percolation in the capillary dominated flow regime. 

It's not really clear what sat_tol does.

The printed messages should really be logger messages but Jeff and I discussed how it would be useful to be able to direct certain messages to a log file so this is a separate development issue.  Also it would be good to control the reporting, say print every 100 steps or so rather than every single one.

It would be good to have a stopping criteria of simulation time as this is more sensible to a user than number of iterations as the user cannot control the time step and time-step should be printed on each iteration. There also needs to be more warning of how a simulation will proceed before run time.

The assertion that breakthrough is achieved is fine but is this the only test that will validate how the code works? I'd like to see more unit tests.

TomTranter:
Is this pull request superseded by #681 ?

stadelmanma:
@TomTranter As I mentioned before the time stepping logic was taken directly out of the paper. I was recreating the original model written in Fortran because it was going to be even harder to fix than rebuild it. The original model was also limited to a 2-D diamond lattice network. I find it odd that at very small flow rates it runs forever, perhaps the large capillary pressures were making the flow oscillate at the entrance to a throat making the time step exceedingly small.

`sat_tol` was used to set of threshold to "round pores up" (or throats) from filling to filled. For various reasons, rounding, etc. sometimes pores/throats would get very close to being filled but never actually fill because the timesteps would be prohibitively small.  `sat_tol` is not an actual volume but instead the fractional amount filled i.e a value between 0 -> 1.

I actually didn't know the logging module existed when writing this code initially. I found it later on and intended to covert my code over to it but didn't have the time to do so. 

Additional stopping criteria is something I wanted to add right now there is only max_steps and breakthrough. Also something I didn't get around to implementing. 

As above unit tests were something that would have gotten written if I hadn't been shifted back over to working on fracture flow. 

---

As the code stands it is about 60% finished. The core algorithm is done and key methods are implemented. As such it may be better to stick it somewhere outside of the main body of code you maintain and "claim" as OpenPNM. Then it can still exist as a starting point for someone else but you don't have to fix/maintain it.

jgostick:
given that this PR is not complete I think we need to close it.  @stadelmanma I'm so sorry about letting all your hard work go to waste.

stadelmanma:
@jgostick No worries, I wouldn't expect anything unfinished to be merged into OpenPNM. I only wish I had been able to finish it. On the bright side this is hardly a waste. I am not planning on removing my OpenPNM repository so if any students of yours (or anyone else in general) has an interest in IMPES flow simulation they would be able to pick up where I left off.

PR #616, opened 2016-07-28
Dual cubic

jgostick:
This PR adds a CubicDual class to the suite of Network generators.  This class creates a standard cubic, but with an another cubic network interpenetrating it offset by half a lattice cell.  These two dual networks are also connected to each other, which allows exchange of material between them such as heat transfer, or perhaps to model mass transfer from pores centers to walls.  This class is added as the cubic counter part to the DelaunayVoronoiDual class added in PR #604 .


jgostick:
Before merging this I need to make connections between the surface pores of network 1 and 2, to be consistent with how the DelaunyVoronoiDual class works.

Issue #624, opened 2016-08-08
Why not having ``template_circle_shell`` alongside ``template_sphere_shell``?


ma-sadeghi:
Currently, we only have ``template_sphere_shell`` (as far as I know). I think ``template_circle_shell`` (or something similar) could be also as helpful for those who work on 2d networks.

jgostick:
You could trim all pores whose z-coordinates are above and below the R plane?

ma-sadeghi:
It's a computational overkill when it comes to large networks. Suppose you want to create a 2d circular disk with radius = 100 pores. The network will roughly have 31,400 pores (pi*r^2) which should be build instantly. Now, if you first create it using a spherical shell template and then trim it, it would take about 100 seconds since the initial 3d network will roughly have 4,000,000 pores (4/3*pi*r^3).

maghighi:
It's actually easy to support 2d image too. Just 2,3 lines in the 'template' code will give us a circle. I can look into that.

ma-sadeghi:
@maghighi You're right. I use my own, I just wondered why we don't have it as built-in. This should do the job:

```python
def create_circular_mask(radius):
    r"""
    Creates a 2d mask for an arbitrary circular network of the given radius.
    The mask is made of True's and False's for the points inside and outside
    of the arbitrary circle, respectively.
    """
    import numpy as np

    x = np.arange(-radius, radius+1)
    y = np.arange(-radius, radius+1)
    xx, yy = np.meshgrid(x, y)
    dist = np.sqrt(xx**2 + yy**2)
    dist = dist <= radius * 0.99
    
    return dist
```

ma-sadeghi:
@maghighi Anyways, I think having it as a separate function rather than passing an extra parameter (ex. `flag=circle`) to the `template_sphere_shell` is more user-friendly.

maghighi:
Yes. It looks better. Because right now when you send just a scalar, it assumes the image is 3d, so it's better not to change the logic of the sphere. I think probably we need a single method that both "sphere_shell" and "circle_shell" can call behind the scene and these two will be just wrapper methods.

PR #642, opened 2016-09-05
Added a method for circular templates

maghighi:
This PR addresses the issue  #624. Another wrapper method has been added for making circular templates. So "template_sphere_shell" and "template_circle_ring" call one private method in the network 'tools' that produces the image arrays. This PR is compatible with the previous codes. Also unit tests have been added for the new method.


jgostick:
Let's change the name to disk_ring...sounds more technical

PR #647, opened 2016-09-21
Enhanced Drainage algorithm to store total network saturation, same as in OrdinaryPercolation

MichaelHoeh:
I added the lines concerning storing total network saturation into Drainage that are already in OrdinaryPercolation. Will benefit people that want to compute e.g. relative permeabilities using the Drainage Algorithm.


MichaelHoeh:
The PR passed Travis-CI's Python3.4 checks, but Travis-CI failed to run the Python3.5 build it seems. I guess this is beyond my control, the PR should be fine still.

jgostick:
The reason I haven't merged this is that I'm not sure how well it plays with late pore filling.  We are trying to tie up loose ends right now so we can move on to V2.  I'm going to close this PR since it's not a critical feature and I'm not sure it's robust enough.  I will add this idea to the issues so we can capture the fact that it's needed eventually.

PR #649, opened 2016-09-28
Fix cubic dual

jgostick:
There was a major bug in this class (the network size was being hard coded, presumably due to some left over debugging lines).  This could have been avoided it we'd done proper team-based code review before merging.  Also, I took the opportunity to adjust how the duals were generated, so now requesting a shape of [4,4,4] returns a [4,4,4].  It was previously being padded to [5,5,5], which was not very logical.


jgostick:
@maghighi , I have set the master branch to 'protected' which means that all PRs must be approved by someone other than the author(s).  I don't quite know how this feature works, but you have to review the code and give it your OK.

Issue #638, opened 2016-08-31
interleave_data returns non-sense when not all pores have been assigned to a Geometry


jgostick:
If you do:

``` python
pn = op.Network.Cubic(shape=[4,4,4])
geom = op.Geometry.Stick_and_Ball(network=pn, pores=pn.Ps[:-2])
```

Then this array will contain non-sensical values in the last two elements:

``` python
pn['pore.diameter']
```

jgostick:
Options are:
1. issue a warning and then fill missing spaces with NaNs
2. throw an exception of some sort

jgostick:
solved by PR #665

PR #665, opened 2016-12-07
fix the bug in interleave data

jgostick:
This addresses #638 

Added new _interleave_data method that might be 50% shorter than original, is mostly working, and hopefull still fast.


jgostick:
@TomTranter Once you're done celebrating you shiny new doctorate, we need to chat about this PR.  I think there is some funky behavior in the old _interleave_data that the new one does not support.  We need to decide if this old behavior is needed or not.  Basically, one of the docstring tests from the old _interleave_data is failing so we need to either change the docstring or change the new method.

TomTranter:
@jgostick I'll look at it next week if that's ok. If it hasn't broken the voronoi vertices then that's a good start but it will need to be checked thoroughly which I'm happy to do. Also, I know you like re-writing all my code but the original problem could have been solved by filling the arrays with nans or zeros. Probably just a one line change with minimal impact : )

jgostick:
yes, rewritting code is a bad habit of mine.  Your version is actually still present for the precise reason that I wasn't sure if mine would be better or not. I just wanted to see if I could do it more concisely.  Feel free to update yours and scrap mine if that works best.  I only spent a few hours on it, so no big loss.

jgostick:
I should also add that I don't think the fix was that simple :-D  If you can get it done by adding 3 lines or less I'll buy you a pint.  Also, be sure to check the speed of the two.

TomTranter:
Hey Jeff,
I've confirmed that a one line change would have fixed the original problem. Sticking "temp.fill(sp.nan)" just before "for item in sources:". However, using the old method breaks some of your new tests, so there's a difference in functionality. I've updated the tests but it may not be the way you want things to work. Basically you throw an error when data types are different for found arrays with the same name on different objects and I convert the data where possible. That was what the error in the codecov tests was about. Your method is shorter but I don't know if it's much quicker... I'll test that tomorrow if you want but have a look at my little commit and feel free to keep it or get rid and then update your tests.
Cheers

jgostick:
Hey @TomTranter it seems that your fix doesn't work...try this:

```
net = op.Network.Cubic(shape=[5, 5, 5])
geom = op.Geometry.Stick_and_Ball(network=net, pores=[0, 1, 2])
```

If you then do `net['pore.diameter']` the elements 3+ are not NaN...still just arbitrary floats.

jgostick:
I'm going to add a unit test to this PR that will fail if nans are NOT present when the above is performed.  This way we can keep trying the code until is passes.

jgostick:
@TomTranter I added 3 unit tests for ints, floats and bools, but not one for objects.  When the array being interleaved is float it correctly returns nans, but when it's ints it returns gibberish numbers and when it's bools it returns all Trues. There is not such thing as a nan for ints, so that must be the problem there.  The correct behavior for ints should be to convert the array to floats so that we can put nans in the empty spots.  For the bools it should be Falses.  Not sure what about objects...maybe empty lists?  or None?

TomTranter:
Hi @jgostick my feeling is that this problem is too small and unimportant to warrant making any more changes to interleave data. [This thread](http://stackoverflow.com/questions/12708807/numpy-integer-nan) mentions making ints that should be nans as some random negative number which the memory issue already kind of does. We don't really want pores not belong to a geometry and so I think we should add a network health check message along those lines when undefined pores has length greater than zero

jgostick:
Odd, why would there be conflicts in those files??

TomTranter:
Did you just change them too?

TomTranter:
Looks like we both made similar changes to get travis passing on our PRs. I kept mine 
:)

jgostick:
Yes, I fixed them in PR #680 and merged into master.   Also added python
3.6 support.

TomTranter:
@jgostick can you review and merge please

TomTranter:
Oh wait, I'm the reviewer so have done it

PR #671, opened 2017-01-22
Network import returns a Geometry

jgostick:
Adjusting the IO classes to return a Geometry along with the Network

### DIFF
--- a/OpenPNM/Utilities/IO.py
+++ b/OpenPNM/Utilities/IO.py
@@ -35,7 +35,7 @@ class GenericIO():
         return geom
 
     @classmethod
-    def _update_network(cls, network, net):
+    def _update_network(cls, network, net, return_geometry=False):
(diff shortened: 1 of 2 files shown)


jgostick:
I have implemented an approach here that is NOT was we had discussed.  In this PR I have first refactored the code so that there is a GenericIO class which all the other classes inherit from.  This does not change the behavior at all.  But I added a new method this class called `fetch_geometry` which takes a network as an argument, strips out all the `props` (except 'pore.coords', and 'throat.conns'), and put them on a new GenericGeoemtry object, which gets returns.  With this approach it means that the user needs to know this method exists and why to call it.  The original plan was to have a flag that would automatically return both a Network and a Geometry if True, but this means that the classes would sometimes be a network and other time a list containing two objects, which is an inconsistent pain.  

We could consider printing a warning that 'fetch_geometry' should be called?

Issue #672, opened 2017-01-24
Strange behavior by ```net.pores()```


ma-sadeghi:
I just realized ```net.pores('all', mode='not')``` which is expected to return an empty list, returns the list of all pores. Is that on purpose? @jgostick

jgostick:
This was fixed in PR #673

PR #673, opened 2017-01-25
Fix Issue #672

jgostick:
Fixing the error in the pores and throats methods that used a shortcut when label was 'all'


jgostick:
Hey @ma-sadeghi ... as per out Github settings, someone needs to review the changes.  Probably nothing much to look at, it passes all tests.  All I did was to remove the shortcuts that were being used when the label was 'all' which caused the mode to be ignored.  The logic was "If the user wants 'all' why would they use a mode?"  But I agree with you that the method should work consistently.

PR #675, opened 2017-02-04
Fixed strange code in NetworkX import

jgostick:
Fixed potential bug where array was being overwritten on each loop. It now checks and only writes array if it doesn't already exist.  

( I have no idea why or how the old code worked.  It was sloppy coding on my part, but I obviously would have caught it had there been an error or problem. )


jgostick:
This is passing except for a small decrease in code coverage...it should be merged.

PR #676, opened 2017-02-08
Add pyamg solver

jgostick:
Added support for pyamg.  You can can specify 'rs' as the ```iterative_solver``` in ```solve```.  If you don't have pyamg installed it will throw an error.


jgostick:
I don't think we should add this.  We already support people using their own solvers, using ``setup`` and ``solve``, so this is a perfect example of this.  We should only support scipy's solvers.

PR #679, opened 2017-02-14
Add generate_voxel_image function

jgostick:
This PR adds a function to ```Utilities.misc``` called ```generate_voxel_image``` that converts pore and throat sizes into a voxel image, which is useful for direct numerical simulation on the network if one so desires.  It only works for spherical or cubic pores, and cylindrical or cuboid throats.


jgostick:
This function is pretty cool but there are a few questions to resolve:
1. Where should we put it?  Right now it's in Utilities.misc...should it be in IO somewhere?  
2. Should we alter the call signature?  I like the max_size argument, but it's obvious what/why this is.

jgostick:
@TomTranter Can you give a review of this so we can merge it?  I think it should be moved to the network.tools folder for now.  I also like the my max_size argument, but I could be convinced otherwise.

TomTranter:
@jgostick something isn't quite right with this function. The code is also not well documented (docstrings are fine but line by line comments on the logic would be appreciated) so it's hard to understand what's going on. Using the short example in the UtilitiesTest script I get two images but taking a slice through the middle it seems like the throats are much smaller in diameter for the cylinders than for the squares.
image
Also, it seems odd that you would place solid all around the edge. If the point is to do dns on these images they are going to need open boundaries. Maybe there could be an option to place the boundaries of the image at the centers of the boundary pores.
Also, the throats seem a little off center compared with the pores on the grid alignment. I don't know whether that's a rounding error somewhere or a feature of the image size. You would get different results for odd and even numbered domain lengths. And finally can you specify whether pore coordinates are cell centers or not in the docs. 
I don't really like the way you define the domain size rather than the pore resolution. It may be safer but I would rather specify a voxel size and get a memory warning/error.
Personally I would take this out of OpenPNM and put it in PoreSpy.

jgostick:
Thanks for the detailed input...although I like this little function, it's definitely not a 'must have'....I'm closing this PR...sorry to waste your time.

TomTranter:
I like it too. Just needs a bit of work but agreed not a must have.

PR #681, opened 2017-02-16
Viscous drainage

jgostick:
I am experimenting with the forking/PR work flow.  I added @stadelmanma's repo to my desktop repo in SourceTree, then I merged our master into his code, then made PR to our repo.  Seems to be working so far.


jgostick:
@TomTranter Can you take a 5 minute look at Matt's code and make sure I didn't miss anything?  I'll merge it when you're done.

stadelmanma:
@jgostick I haven't had a chance edit the code since we last talked so version you forked might be a little rough around the edges but it works. As before when running the code regular cubic networks won't work well because pores will essentially fight over shared throats at the inlet. I'll be glad to answer any other questions you or the OpenPNM team might have regarding the code, I think I still get notifications from gitter if my name is mentioned.

Additionally I rebased my master branch off yours and then merged it into my branch of viscous drainage just incase. Somewhere along the way awhile back I broke something and had to fix a bunch of conflicts. It seems like your method worked and likely will end up with a cleaner overall history though.

jgostick:
Hi @stadelmanma, @TomTranter and I had a look at the code, and we have a few concerns/issues:

1.  Does it really only work on a 2D diamond lattice?  I assumed it would be more general than that.  If this is true, why is it so, and can it be generalized?

2.  Why does it get stuck?  Is there not some way to detect this and adjust things?

3.  It seems to be quite slow.  How long does it take for you, and what is the maximum network size you've used?  Is it slow because Python?

4.  Would you be willing to make an example for the examples repo that shows how to reproduce some of the figures from Ferer's paper(s)?  Is there a way to get simulations in all 4 corners of the capillary phase diagram?

5.  How flexible is this for different pore and throat sizes?  Your integration test have very small pores and long throats.  What about large pores and stubby throats?

6.  If you run your simulation at creeping flow rates does it reproduce the standard invasion percolation exactly?  Is should, so maybe this could be an other test to add.

stadelmanma:
I'll answer all these the best I can since I haven't run it in awhile. 
1. It can run on any pore network in theory. I use a 2-D diamond lattice for testing because it represents the simplest network possible that also covers all of the algorithm’s behaviour. 

2. It gets stuck on simple cubic networks because two pores are both trying to advance into the same throat. Instead of one pore "taking over" essentially what happens is one pore pushes into the throat pushing the other's meniscus out. The next time step the other pore pushes it's way back in and the other pore out. I have to limit my time steps to certain criteria to maintain mass balance (and logic), one of which is based on a meniscus entering or exiting a throat. When two pores fight over a throat the time steps get very small and essentially nothing happens. 
    * A solution could be to perturb the system by a much larger amount when enough movement isn't detected (or too small a timestep). I don't know how that will effect things like mass balance. 

3.  It's largely slow because for an IMPES simulation to be somewhat valid you can only allow for very small time steps. I based my numbers off of those used in the paper, since I was essentially recreating their model. If you allowed a larger amount of throat movement when capillary pressure is increasing say change it from 3% to 5% or even 10% then you might notice a bigger speed up. I don't know how that would effect the simulation, I have a feeling if a large delta was allowed your capillary pressure changes might introduce some odd behaviour. The second reason it is slow is likely because it's python. For small networks `_advance_interface` and  the linear solver take similar amounts of time. I've optimized `_advance_interface` as best I could. You also have to keep in mind it calls that linear solver a every single time step, unlike say StokesFlow which calls it once.

4. I can probably do that but it likely won't be until April. I'm currently juggling a full time job and a Master's Thesis that I procrastinated a little too much.  

5. It will work for any network however as I mentioned above the main limiting factor is the restriction on how much a meniscus can advance while capillary pressure is increasing. Pores can fill up to 25% per time step but the time step is set by the smallest possible advance in the entire network. So in situation with large pores and small throats your pores will take a long time to fill. That was an issue I had running it on a real network from our CT data. The test is setup to run fast and large throats help speed things along.

6. I made that comparison on a few networks and it wasn't exact but it was very close. I may have had too high an injection rate.

stadelmanma:
@jgostick you are more than welcome to close this and #607 I am working outside of the research field now and I won't be able to return to this work (at least not in any realistic timeframe). I apologize for throwing something half-finished on your doorstep, I wasn't expecting to get pulled away from this work right in the middle of it when I did.

I was able to take away a lot of useful Python programming skills from working with OpenPNM (proper logging, packaging, CI, read the docs, etc.) that have helped me immensely. Thank you for that. If there are any students interested in viscous flow through PNs you are more than welcome pass this code off to them as a starting point and I'll be happy to answer any questions they have.

PR #694, opened 2017-04-12
Make OpenPNM Great Again!

ma-sadeghi:
Optimizing OpenPNM to run faster.


jgostick:
Wasn't there a few more speed-ups you were going to do, like stop checking network health on each merge, and something about not building adjacency_matrices each time as well?

PR #706, opened 2017-06-26
name objects sequentially rather than randomly

jgostick:
Changed the way objects are named, to be numbered sequentially, rather than assigned random names.  Should make paraview much easier to use.

### DIFF
--- a/OpenPNM/Algorithms/__Drainage__.py
+++ b/OpenPNM/Algorithms/__Drainage__.py
@@ -574,7 +574,9 @@ class Drainage(GenericAlgorithm):
             temp_Pc = phys.models[key]['Pc']  # Store old Pc
             phys.models[key]['Pc'] = pressure
             # Regenerate Physics model and capture output locally
-            locs = self._net._get_indices(element=element, labels=phys.name)
+            locs = phys._map(element=element,
+                             locations=phys._get_indices(element),
+                             target=self._net)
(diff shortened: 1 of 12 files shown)


jgostick:
This is a good idea and can be done with almost no backwards compatibility issue...except one problem: algorithms are not registered with the workspace manager when they are created.  The sequential object numbers are derived from the objects found the workspace manager dictionary, so since algorithms are not there they numbering gets screwed up.  The main repercussion is that we could end up with two algorithm objects named 'drainage_0007' for example.  This would occur if seven objects were already created and then we created two drainage algorithms in sequence.  The number of objects in the workspcae will not have increased.

jgostick:
I guess the question the reviewer has to answer is whether or not we should list algorithms in the workspace object (I think we should).  If yes, then there is a bit more work to do on this.

jgostick:
In the spirit of improving the naming thing, we could also allow duplicate names by appending a number to them if the name already exists and throwing a warning, such as:

``` python
    pn = op.Network.Cubic(name='bob')
    pn2 = op.Network.Cubic(name='bob')
    Warning: An object named 'bob' already exists, using 'bob_001' instead.
```

jgostick:
I'm going to close this PR, and create it as an issue to be dealt with in V2.0.  We need to tie up loose ends as much as possible right now.

PR #708, opened 2017-06-27
adding monkeytest.it to the travis.yml file.

jgostick:
After inadvertently breaking the links in the docs, I looked and found this service that will test the html including searching for broken urls


jgostick:
this did not work as advertized

PR #709, opened 2017-06-29
Fixing Numpy boolean compatibility

maghighi:
Rearranged some of the boolean masks in the linear transport algorithm to make the code compatible with the new version of Numpy.


jgostick:
There are still two errors:

ValueError: shape mismatch: value array of shape (46250,) could not be broadcast to indexing result of shape (16875,)

This is better, you fixed one error, but one more to go.

maghighi:
The error of ''shape (1903,3) could not be broadcast to indexing result of shape (1903,)'' is related to the line 1104 of the file Utilities/IO (class iMorph, def load).
Since I'm not familiar with this class, I ask @jgostick to fix it.

PR #711, opened 2017-07-05
Adding version / tag comparison check

jgostick:
Adding new file called PackageTest that contains general tests about so far this only ensures the version and the latest git tag match


jgostick:
this is fun idea, but it's not really working.  I'm going to close it.

PR #714, opened 2017-08-05
added a function to network.tools for labeling faces (top, bottom) etc

jgostick:
Called label_faces.


jgostick:
This PR is good to go...it's just a simple function added to network.tools.  It does the same as find_surface_pores, but does it once for each face and adds 'left' and 'right'

PR #730, opened 2017-09-12
Read networkx graph object directly

joosthvanderlinden:
As per personal communication with @jgostick , `NetworkX.load()` is replaced by a function that accepts a `networkX` graph directly, instead of using an intermediate YAML file.

Checks on the input graph:
- Checking if supplied object is networkX graph
- Checking if supplied graph is undirected
- Checking if numbering of G.nodes() contains any gaps
- Checking if numbering of G.nodes() starts at 0

The implementation is very similar to the original YAML approach. There was no need to remove duplicates from `conns` anymore, but it's still sorting because I could not determine if `G.edges()` returns a sorted list.

I didn't manage to fill an `ndarray` of strings, such as `_sp.ndarray((N,M), dtype=str)`, correctly. hence there are a few checks to see if `dtype` is `str`. Without these checks and the use of `dtype='object'`, the `ndarray` would contain only empty strings. Perhaps there is a more elegant way to handle this?

Lastly, `test_load_networkx` in `IOTest.py` will need to be updated.

### DIFF
--- a/OpenPNM/Utilities/IO.py
+++ b/OpenPNM/Utilities/IO.py
@@ -1023,7 +1023,7 @@ class NetworkX(GenericIO):
                     val = {i: phase[prop][i] for i in network.Ps}
                     _nx.set_node_attributes(G, prop[5:], val)
                 if 'throat.' in prop:
-                    val = {tuple(conn): phase[prop][i] for i, conn in \
+                    val = {tuple(conn): phase[prop][i] for i, conn in
                            enumerate(conns)}
                     _nx.set_edge_attributes(G, prop[7:], val)
         return G
--- a/OpenPNM/Utilities/IO.py
+++ b/OpenPNM/Utilities/IO.py
@@ -838,7 +838,7 @@ class NetworkX(GenericIO):
 
     2. Since \'pore.coords\' is so central to OpenPNM it should be specified
     in the NetworkX object as \'coords\', and the [X, Y, Z] coordinates of
-    each node should be a 3x1 list.
+    each node should be a 1x3 list.
 
     3. Edges in a NetworkX object are accessed using the index numbers of the
     two nodes it connects, such as ``G.edge[2][3]['length'] = 0.1``
@@ -860,9 +860,9 @@ class NetworkX(GenericIO):
         Parameters
         ----------
         G : networkx.classes.graph.Graph Object
-            The NetworkX graph. G should be undirected. The numbering of nodes 
-            should be numeric (int's), zero-based and should not contain any 
-            gaps, i.e. ``G.nodes() = [0,1,3,4,5]`` is not allowed and should be 
+            The NetworkX graph. G should be undirected. The numbering of nodes
+            should be numeric (int's), zero-based and should not contain any
+            gaps, i.e. ``G.nodes() = [0,1,3,4,5]`` is not allowed and should be
             mapped to ``G.nodes() = [0,1,2,3,4]``.
 
         network : OpenPNM Network Object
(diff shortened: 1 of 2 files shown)


jgostick:
Many thanks for doing this.  I hope I can get to it before you return from your break!

jgostick:
Hey @joosthvanderlinden I added a branch to our repo and changed the target of your PR to that...I 'think' this will allow me to edit your code (I've done very few PRs so I'm not totally sure about this).  Anyway, it'd be good if you could fix the errors that are occurring in your code....the I can look at it and play around with it.  We also might add an 'export to networkX object' function to this PR, right @ma-sadeghi ?

ma-sadeghi:
@jgostick Yes. I recall that we agreed to call `networkx` directly to do the export to `yaml`, right? I'll copy the code snippet I used to create `networkx` objects.

jgostick:
@ma-sadeghi Excellent, let's roll these two efforts in to one.   I have created a branch and PR for this, so you can tack yours on there. Also, we should create a new class called NetworkX, instead of YAML.  We can keep the YAML one, but need to make it more robust to handle variations in the in coming file.

joosthvanderlinden:
I updated the unit test `test_load_networkx` so the code should pass now.

Can you clarify the following?

> I have created a branch and PR for this, so you can tack yours on there.

Should I move my changes somewhere else? How do I go about that?

> Also, we should create a new class called NetworkX, instead of YAML.

The existing class was already called `NetworkX`. I replaced all YAML-related code with a method that directly reads in a networkX graph. A `save()` method in this class doesn't exist yet, so @ma-sadeghi could simply add this.

joosthvanderlinden:
Additionally, NetworkX [just released version 2.0](https://networkx.github.io/documentation/latest/release/release_dev.html) and there are some [major changes](https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html) that may break the current PR. Would you prefer to stick with < 2.0 (and make that a requirement somehow) or convert this PR (as well as exporting to networkX) to networkX 2.0?

jgostick:
Interesting... perhaps we should support both for a while so either have a
separate class for each like NetworkX_V1 or have an argument somewhere that
specifies the version.

joosthvanderlinden:
That might work. Is it common for a library to support multiple versions of an external library like that? i.e. is it good practice do check something like 
```python
import networkx as nx
if (nx.__version__ < 2.0): #this will return a string though
    #do something
else:
    #do something
```

joosthvanderlinden:
Any thoughts @jgostick ? I'd happy to implement something that works well for you.

jgostick:
This PR looks amazing!  Sorry I let it drag on.  Do you have a feel for how different V2 of NetworkX is?  How much extra code will be involved to support both?  Perhaps we should write a second class called NetworkX2?  Maybe we should add your above code but more like:

``` python
if (nx.__version__ >= 2):
    raise Exception('NetworkX V2 is not supported yet')
```

Then we can eventually change that Exception to instead call the NetworkX2 class?

jgostick:
Hey @ma-sadeghi  , how close are you to getting your NetworkX writer onto this PR?  @joosthvanderlinden has put a lot of work into his 'read' part, so I'd like to merge it soon.  Ideally this class should also contain the 'write' part.

ma-sadeghi:
@jgostick Sorry for the delay. It's done now. The `save` method exports OpenPNM Network object to NetworkX. In case we're still interested in importing/exporting from/to actual files, we can use NetworkX's `write_yaml` to export to `yaml`. @joosthvanderlinden can then read the `yaml` file using NetworkX's `read_yaml` and then pass the imported NetworkX object to the `load` method he implemented.

jgostick:
I pushed a change to the travis.yml file so that python 3.4 stops getting tested, but my pushes are not showing up in the PR for some reason, even though they are in the code.  This is also happening on my other PR in porespy.  Will investigate.

jgostick:
This PR is vexing me!  I think the problem stems from my PR'ing from Joost's fork.  I don't seem to have @joosthvanderlinden 's  code in my branch, I only see @ma-sadeghi .

ma-sadeghi:
@jgostick I had the same problem, so I push my code to Joost's fork. I think you need to merge his branch into OpenPNM's main repo.

PR #737, opened 2017-10-20
Improved networkx compatibility

jgostick:
This Fork and PR stuff if very confusing....I already did this once, and we had a long conversation that is now gone.  Oh well, this PR seems to have both @ma-sadeghi and Joost's code


ma-sadeghi:
@jgostick I see some other changes irrelevant to this PR in "Files changed".

jgostick:
damn..I switch git clients and it apparently pushed ALL my files!  >:-)

jgostick:
Hi @joosthvanderlinden and @ma-sadeghi.  This build is not working...and the error output on Travis is very hard to follow.  I don't know which parts of the code are breaking, but would you both mind taking a look?

ma-sadeghi:
@jgostick It's weird. I just ran all the tests on my machine and they all passed!
image

jgostick:
I'm guess that you didn't include the doc tests in your tests?  Looking at the Travis error makes me think it is the examples in the doc strings that are breaking.

ma-sadeghi:
@jgostick Actually I think it's the new NetworkX because, on my machine, I have the old one (v 1.11 I guess). I'm confused now if we were supposed to support the latest version or not?!

jgostick:
I also don't remember what we decided :-)  What do you think?

ma-sadeghi:
@jgostick Maybe support both for a while and then drop the support?

jgostick:
@ma-sadeghi , it seems that the failure is coming from your save function....Travis is using NetworkX V2, and I guess your code was designed for V1.11?  @joosthvanderlinden 's code works, because he specifically instigated this PR to make these IO functions compatible with V2.  I guess we should support both version, but I'm not sure the best way....perhaps with two classes, NetworkX_V1 and NetworkX_V2?  or we could put the version number as an argument?

ma-sadeghi:
@jgostick All three tests are failing including my two tests on save with and without phase as well as Joost's test on load. So I guess neither mine nor Joost's code is compatible with the latest version. I haven't looked into the compatibility issue yet, but it seems that it's a naming convention or something of that sort. If there's a way to know the version of a NetworkX graph object, then we can handle both versions with some if statements, at least for a while until we're sure that nobody is using NetworkX v.1.X anymore.

jgostick:
The NetworkX website has a [guide for transitioning to V2](https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html) that gives instructions on how to support both.  I will try this.

test/unit/Utilities/IOTest.py
```diff
@@ -144,15 +145,42 @@ def test_save_and_load_mat_w_phases(self):
...
+
+    def test_save_and_load_networkx_w_phases(self):
+        G = io.NetworkX.save(network=self.net, phases=self.phase)
+        net = io.NetworkX.load(G)
+        assert net.Np == 27
+        assert net.Nt == 54
+        assert sp.shape(net['pore.coords']) == (27, 3)
+        assert sp.shape(net['throat.conns']) == (54, 2)
+        assert [True for item in net.keys() if 'temperature' in item]
+        assert [True for item in net.keys() if 'diffusive_conductance' in item]
```
joosthvanderlinden: This assertion failed for me - 'diffusive_conductance' is not in `net.keys()`. Easy fix?

PR #744, opened 2017-11-21
Change to way csgraph is imported. Old way was broken by Scipy 1.0

TomTranter:


TomTranter:
I've checked to confirm whether this change is backwards compatible and it is.

PR #764, opened 2018-01-29
MixedPercolation

TomTranter:


TomTranter:
@jgostick This is ready to be merged. Would you like to run through it together before doing so?

TomTranter:
@jgostick what do you want to do with this code?

PR #775, opened 2018-02-01
Added dispersion algorithm (advection + Fickian diffusion).

ma-sadeghi:


ma-sadeghi:
The Dispersion class implements advection diffusion based on the exponential upwinding scheme. Note that although it subclasses the `GenericLinearTransport`, because of the advection term (which makes the `A` matrix asymmetric), it has its own `build_A` method, which overrides that of its superclass.

jgostick:
The build_A matrix surely works, but it is not vectorized at all....isn't it terribly slow?  You'll have to walk us through why you need such a complicated pore by pore approach.

magnaou:
I have pushed new advection diffusion solver (discussed with Jeff and during March's group meeting).
The equations solved come from Mehmani et al 2014, 2017 works.

PR #789, opened 2018-02-05
Refactor network queries

jgostick:
This pulls the network queries out of GenericNetwork and into topotools.  Even if this get abandoned, there are a few useful bug fixes and speed-ups that should be kept.


jgostick:
As of my last push, all the network queries are moved to topotools and GenericNetwork only wraps them.  This means we can eventually pull them out of topotools into a more general package.  I think I'll merge it now

PR #830, opened 2018-03-19
advection diffusion solver plus an example script

magnaou:
A new "build_A" method is implemented in this solver. 
I couldn't use the existing "build_A" method because in the advection diffusion solver the diagonal terms of the "A" matrix are computed from a conductance array that is different from the one used for the off diagonal terms.
Mehrez


jgostick:
@magnaou welcome to the annoying world of pep8 errors...Travis enforces the pep8 formatting, so you need to get everything just right.  The main error I'm seeing is that comments must have a space after the # sign, and if they follow code, there must be 2 spaces before the #.  You should turn on linting in spyder so you get the little yellow warning sign for these errors.

magnaou:
@jgostick Thank you for the comments. Fixed, it should work now.

jgostick:
this code is on dev-V2-alpha1_merhez, along with amins, for sidebyside comparison, so this will be closed.

ma-sadeghi:
@jgostick Are you sure? I don't see any other branch.

jgostick:
NO, I screwed this up.  @magnaou is going to repush his code

jgostick:
Oh man, now I've really screwed it up!

PR #852, opened 2018-04-12
Updating percolaton

jgostick:


TomTranter:
@jgostick - I can tell you are going to hate some of this code for MixedPercolation - there's a lot of it and it's not all easy to understand or in the right place. I want to reopen the issue of models returning multiple bits of data and go one step further and introduce model classes. I have a Purcell and Sinusoidal capillary pressure model and they are both capable of calculating the threshold entry pressure but also meniscus properties for a given pressure such as filling angle meniscus radius and center position. Having these all as functions or in one big function inside a capillary pressure model seems like the wrong thing to do. Maybe they should actually form part of an interface class. It would enable me to also add some additional functionality for plotting which was really useful for debugging - but I did it separately in an ipython notebook then copied the code over when I was happy.

so I guess you could write something like

inter = op.interfaces.toroid()
phys.add_model('throat.capillary_pressure', inter.max_entry())

Could be overkill but it feels like too much code to go in a pore-scale model at the moment

jgostick:
So you're getting the *let's make V2 perfect* bug just like me!  Interface classes make a lot of sense...and we should all talk about them together for an hour at least.  I can't do the telecon tomorrow, but you and I can start to chat about it, then bring to everyone next week.  Adding an Interface class sounds like a version 2.1 thing though.  I really want to get 2.0 functioning and on Pypi for download by May 12 (interpore).

TomTranter:
Fair enough. I will focus on just getting this thing working then

TomTranter:
The tests associated with MixedPerc are now passing but I'd like to improve them. The Drainage ones are now failing because some keywords are unexpected. I'm not really sure what's going on with settings and kwargs etc. in V2. Can you go through this in the next meeting too. I'll make an agenda for next Tuesday

TomTranter:
Is the format for the percolation class structure pretty much set now? I want to know before I start changing things around with MixedPercolation to try to fit it in.

TomTranter:
@jgostick - quick question on outlets. Funny we were talking about how you wanted to do this earlier. I actually made MixedPercolation accept an outlets argument for trapping but also as part of the run method to terminate the cluster. This is because for GDLs (and maybe in general) once a percolating pathway has established the resistance at the outlet is less than the resistance to invade new pores connected to the cluster so invasion basically stops growing, Should this be another setting or default behavior?

TomTranter:
I guess outlets for trapping and cluster termination will always be the same. I should also check carefully how these two things work together i.e. can you successfully post-process trapping if it has only partially saturated the domain

TomTranter:
Should we make setting outlets automatically terminate clusters and apply trapping or control them individually?

TomTranter:
@jgostick I had a thought on the multiple clusters. The way it works now is basically to grow one element at a time whether that is a pore or throat based on the lowest connected entry pressure. The highest capillary pressure required to invade up to that point in time is stored as the invasion pressure and the invasion sequence is incremented for each cluster once an invasion is allowed. If clusters meet which happens when an element is invaded by a cluster and it has already been invaded by a different one they are merged and one queue is emptied into the other. This does not necessarily guarantee that both clusters are at the same pressure when they merge as they could have invaded into the same pore through different throats and so if reproducing an experiment controlling percolation by varying pressure it is not realistic. Also, if controlling by mass flow the sequence does not necessarily correspond to the same invasion time as the volume in each cluster at a given sequence will almost certainly be different. We need to think again about how this algorithm is progressed from one step to the next

TomTranter:
@jgostick passing now on coverage but only cos I shoe-horned in a few tests on vertexops. What's your plan for SiteAndBond? Currently has no dedicated test. Also GenericPercolation is quite low coverage.

TomTranter:
@jgostick a question on clusters and snap-off. If multiple clusters are allowed and they are allowed to invade separately then this means they can be at different pressures at the same time. For injection from multiple sources this makes sense to me. However for imbibition, snap-off of the wetting phase should be at the same pressure everywhere and if a throat snaps this will start a new cluster which can then go on invading at it's lowest pressure point. So there is inconsistency. I guess access limited invasion can be at different pressures and have simultaneously growing clusters but access unlimited should be treated as one cluster. Do you agree?

PR #855, opened 2018-04-12
Sympy source terms

TomTranter:


TomTranter:
@jgostick at the moment I have not removed the old methods and have the new ones running alongside with tests checking that they produce the same results. They are passing. We could remove all but the general_symbolic method or keep the old functions but have them as wrappers to general symbolic. Documentation would need to be improved and an example with analytical solution would also be good to have if we did remove the old ones. I don't know how often people really use all the different functions

jgostick:
A few comments: 
* Why does the ``arg_map`` have to be an OrderedDict?  Why not just a dict?  Like just
arg_map={'a': 'pore.a', 'b': 'pore.a', 'c': 'pore.A3', 'var': 'pore.x'}
* Line 609 says if 'x' not in keys, but 'x' is stored in 'arg'...I'm getting an error that I feel I shouldn't get
* Are people meant to call ``build_func``?  It looks like it should be private
* all your func accept kwargs, but openpnm doesn't need this any more
* the docstrings could use some beefing up...usually you just need to write one good one then cut and paste.  It seems redudant, but users only see the doc string on the func they're looking at.  (hmmm, I wonder if docstrings can be subclassed? :-) )

jgostick:
when I pass ``var = 'pore.x'`` I get an error that var was not passed, and when I pass ``x = 'pore.x'`` I get an error from the ``build_func`` function

TomTranter:
@jgostick Wasn't sure what you meant by the line 609 comment. All others addressed now

jgostick:
The example helps a lot!

Only issue is that I was imagining the user passing in the function as a string, like eqn='a*x**b', and that sympy would parse this into an equation.  This way the user doesn't have to know how sympy works, we'd just pass their equation to sympy inside the model.  Is this a bad idea?

TomTranter:
There is parsing but to be honest I'd rather not use it. The basic sympy implementation we have is easy enough to use and it forces people to use the sympy versions of more complex math functions like sin and log. Keeping things simple on our side also reduces chances of bugs and coding overhead

PR #857, opened 2018-04-17
Use Cantera to calculate physical properties.

ma-sadeghi:
Following the discussion we had a few weeks ago, this purpose of this pull request is to add `Cantera` as a dependency to `openpnm` for calculating physical properties such as thermal conductivity, diffusivity, etc.

So far, the major downside that prevented us to do this is that it takes `cantera` roughly 10 times compared to directly calculating properties with our current pore-scale models. The reason is probably that `cantera` updates all the physical properties attached to a phase even when you only need one.


jgostick:
this needs to be re-PRd onto dev

PR #859, opened 2018-04-18
XDMF now uses version 3 and throats are cell type, not node

jgostick:


jgostick:
One fairly major change is that the hdf5 file now uses actual nested data structures to store data, instead of a flat dict.  Paraview seemed to dislike the use of pipe (|) delimiters, so I just decided to use normal hdf5 format...the naming in paraview still uses the pipe though, so it's nice and clean

jgostick:
I was thinking of leaving this open until I found a way to add 'throat vector' information, but I think that can wait...so close and merge.

PR #888, opened 2018-05-02
New dependency tree

TomTranter:
Dependency tree with error for circular dependencies


TomTranter:
hey @jgostick please approve the change to max size

ma-sadeghi:
@TomTranter Sorry for intruding on your PR. I added `dependecy_tree2()` that's built on top of `networkx`. Might be worth considering. Apparently, what we want from `dependecy_tree` is called [topological sorting](https://en.wikipedia.org/wiki/Topological_sorting) in graphs, and `networkx` already has it.

TomTranter:
@jgostick I take it you're half way through changing something on this branch to do with ModelWrapper as it's making everything fail... I'll leave this PR to you to finish

jgostick:
I was working on this too late last night, and didn't realized I left so code in there from another thing I was working.  Anyway, now it's working.  I made a few tweaks, hopefully for the better:
1. I changed the name to dependency_list, since it returns a list after all (and I updated your nice tests accordingly)
2.  I added dependency_graph which returns the networkx object, and dependency_map, which attempts to plot the graph for visualization.
3. I changed the networkx function so that it returns a unique list (it basically does the topological sort, but uses alphabetic sorting to workout ambiguous order).

jgostick:
Hmm, these tests passed locally, so maybe the sort isn't actually working and I just got lucky...will investigate.

TomTranter:
@jgostick @ma-sadeghi we need to think about models that lookup information from objects other than the target. The simplest capillary pressure model takes info from phase and geometry. This would be possible if we implemented models that knew what object they have been attached to and passed model objects rather than dict keys to the models as parameters. So at the moment dependency tree for a given object e.g. phys_air only looks at arguments pertaining to data that is also on that object. This is a useful but rather simple tool. We should keep in mind how to design it to cover the entire simulation. Like if a geometry property changes such as pore volume due to some surface adsorption the tree could Identify all the models using pore.volume on all the physics associated with that geometry

PR #897, opened 2018-05-09
Model wrapper

jgostick:
This adds a ModelWrapper class which houses a model and all it's arguments.  It has the ability to find its own target, and to call itself, so you could do something like this:

```
mod = geom.models['pore.diameter']
mod.run()
```

I am not sure this is worth the complication of adding another level of classes, but it's something we can add at a later date.  

Of interest on this PR is the ability for the ModelsDict (the dictionary that houses the ModelWrapper) to look up it's own master...which may be worth cherry picking.

### DIFF
--- a/openpnm/core/ModelsMixin.py
+++ b/openpnm/core/ModelsMixin.py
@@ -5,6 +5,40 @@ ws = Workspace()
 logger = logging.getLogger()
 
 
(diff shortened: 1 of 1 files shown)


jgostick:
I'm closing this....but I won't delete the branch for a while in case we decide to revisit it.

PR #910, opened 2018-06-08
Better project

jgostick:
Updating and improving all the save and load functions in Workspace and Project.  This include accepting full paths in the filename argument, and allowing loading of simple pickled lists as Projects.


jgostick:
I'm going to merge this so that nivan can start using it in the gui.

PR #912, opened 2018-06-08
Throat endpoints

jgostick:
This is a first crack at tracking the end points of throats.  

Advantages are:
1. Allowing visualization in paraview (since the throat vector data would be known)
2. More accurate estimation of throat length and volume


xu-kai-xu:
@jgostick  hi, when I read the docstrings in the file .\models\geometry\throat_endpoints.py , I found one line (153) with the notation **# Handle throats w/ overlapping pores**.  I don't know what "w/" means. Is this a typo or with some other meaning? Also, I don't know the equation in line 160. So I have to ask for help. May you check it? Thanks!

ma-sadeghi:
@xu-kai-xu "w/" is shorthand for "with"

Issue #909, opened 2018-06-08
Finally a NumFocus-supported package for unit manipulations in Python!


ma-sadeghi:
The yt-project team (part of NumFocus) have recently released `unyt` [package](https://github.com/yt-project/unyt) for manipulating units in Python. The package fully supports `numpy` arrays, and might be worth considering in `OpenPNM`. They've also submitted a [paper](https://github.com/yt-project/unyt/blob/master/paper/paper.md) to JOSS discussing the performance overhead of their package.

jgostick:
I've got a PR (#914) going that shows this in action.

jgostick:
I'm going to close this issue, since there is a PR already open on the matter where further discussion should occur.

PR #914, opened 2018-06-10
Adding tentative support for the unyt package

jgostick:
A first attempt at supporting units through the unyt package.  Related to issue #909 .

### DIFF
--- a/example_script.py
+++ b/example_script.py
@@ -2,9 +2,13 @@ import openpnm as op
 ws = op.Workspace()
 proj = ws.new_project()
 pn = op.network.Cubic(shape=[5, 5, 5], project=proj)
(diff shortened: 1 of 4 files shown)


jgostick:
More tests should be added that really push this to the limit, such as actually adding units to every property, running pore scale models, making sure they all add up, performing simulations and so on.

Also, the current compatibility is entirely behind the scenes, but we might want to consider including 'units' as an optional argument to pore scale models.  The model could either make sure the end result has the given units, or could add the units..  This would be an outward facing change though.

IF we were going to add this feature, then I feel like it might be a good V2.1 or 2.2 item, along with cantera support for calculating properties.

jgostick:
this needs to be re-PR'd onto dev

PR #929, opened 2018-06-21
Working on issues

jgostick:
Trying to fix many of the little issues and bugs leading up to V2 release


jgostick:
I'm going to merge this since it might be a while before we get to the other issues I wanted to fix, since they require discussion among the group and some decisions.

PR #935, opened 2018-06-27
Improved models for calculating equivalent area for conduits

ma-sadeghi:
Previously, we used to use the whole pore area for calculating the transport conductance for a conduit, whereas only part of the area contributes to the total conductance. This PR addresses this issue by introducing a new parameter, i.e. equivalent area, defined for individual conduits such that it encapsulates the variation of cross-section area of the two pores and possibly throats (not yet implemented), which can later be used by conductance models to give more accurate conductance values.

### DIFF
--- a/openpnm/models/geometry/throat_area.py
+++ b/openpnm/models/geometry/throat_area.py
@@ -1,4 +1,5 @@
 from numpy import pi as _pi
+import numpy as _np
 
 
 def cylinder(target, throat_diameter='throat.diameter'):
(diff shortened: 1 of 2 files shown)


ma-sadeghi:
Hey @Zohaib-Atiq, feel free to add your 4-sided pyramid to this branch. Meanwhile, I'll add a model to handle the 2d case.

ma-sadeghi:
Some corner cases that need to be addressed:
1. How do we address the case where a throat diameter is larger than the diameter of either of its adjacent pores?
2. How do we address conduits with overlapping pores?

PR #939, opened 2018-07-02
change A.dtype to np.int64

thirtywang:
Fix Memory limit for scipy.sparse.linalg.spsolve with scikit-umfpack. 
When using umfpack to solve Ax=b, which have 10X speedup in Ubuntu and MacOS, an int32 indices will hit a 32 bit memory limit (such as more than 180000 pores). This is because the 64-bit umfpack family solver is only called when the indices have dtype int64 (see more: https://github.com/scipy/scipy/issues/8278). 
An easy way to fix this bug is to change the dtype of A to np.int64

### DIFF
--- a/OpenPNM/Algorithms/__GenericLinearTransport__.py
+++ b/OpenPNM/Algorithms/__GenericLinearTransport__.py
@@ -6,6 +6,7 @@ module __GenericLinearTransport__: Class for solving linear transport processes
 
 """
 import scipy as sp
+import  numpy as np
 import scipy.sparse as sprs
 import scipy.sparse.linalg as sprslin
 from OpenPNM.Algorithms import GenericAlgorithm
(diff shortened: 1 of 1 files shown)


jgostick:
Very nice to know that such a simple fix!  Your PR is failing because numpy is used in the example script as np, but it not imported.  Can you fix?

thirtywang:
Sorry for forgetting the import. My IDE don't highlight change in Lib.  I think a more robust approach for this problem is to modify the  _build_coefficient_matrix function, which too hard for me.

thirtywang:
Oh, failed again. but I run the **example_script.py** in my local machine (win 10) with scipy 1.0 1.01 1.1 and they all passed. 
I think it beyond my ability.

jgostick:
I'm going to close this PR because we're not really planning to develop V1 any longer.  We are within days of finalizing the V2 release, so I will investigate making this correction to the V2 branch.  

I want to thank you offering this PR and I'm sorry to close it!

jgostick:
I have now implemented this in V2...it works!  Thanks very much.

Issue #945, opened 2018-07-04
Trim not removing correct data


TomTranter:
Currently in the dev-V2 branch the following code results in a True statement and should be False.

import openpnm as op
import scipy as sp
sp.random.seed(1)
pn = op.network.Cubic(shape=[2,2,2], spacing=1)
Ps = pn.pores()[:4]
Ts = pn.find_neighbor_pores(pores=Ps, mode='union')
geo1 = op.geometry.GenericGeometry(network=pn, pores=Ps, throats=Ts)
Ps = pn.pores()[4:]
Ts = pn.find_neighbor_pores(pores=Ps, mode='union')
geo2 = op.geometry.GenericGeometry(network=pn, pores=Ps, throats=Ts)
geo1['pore.random']=sp.random.random(geo1.Np)
geo2['pore.random']=sp.random.random(geo2.Np)
trimmers = pn['pore.random']<0.25
op.topotools.trim(pn, pores=pn.pores()[trimmers])
print(sp.any(pn['pore.random']<0.25))

This is because trim relies on mapping geom _id to net _id and this mapping fails if the network is trimmed first. This should be fixed if the network is trimmed after the geom but requires better testing with more objects. It is possible that this bug has existed for some time but was found when running a Voronoi test that failed to remove zero area throats.

TomTranter:
project does have a reverse method as it's basically a list so the network may not always be at the start of the list.

jgostick:
so this issue is solved then? that was fast!

PR #947, opened 2018-07-04
Fix trim

TomTranter:
Fix addresses issue #945


TomTranter:
@jgostick  I think this fix should be ok providing the network is always at the top of the project list and gets moved to the end when trimming. I don't know what point this bug came in but would happen I think whether it was id or uuid.

PR #952, opened 2018-07-05
Change VoronoiFibers to be project rather than network and update tests

TomTranter:


jgostick:
than of the project is only None when it's not in any workspace...so the problem might be deeper

TomTranter:
Nah, I was making it None in the init of the materials class. Then using None again later for the prefix of the net and geom names rather than the sim_01 that gets generated. It's all good

PR #978, opened 2018-07-11
Minor tweak to use 64 bit integers in A matrix

jgostick:
This was added to force scipy to use the 64 bit version of umfpack and hence increase the allowed memory.  It works!

### DIFF
--- a/openpnm/algorithms/GenericTransport.py
+++ b/openpnm/algorithms/GenericTransport.py
@@ -435,7 +435,10 @@ class GenericTransport(GenericAlgorithm):
             del(ls)  # Clean
         else:
             solver = getattr(sprs.linalg, self.settings['solver'])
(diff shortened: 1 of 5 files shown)


jgostick:
Interesting!  We do not actually include scikit-umfpack in our deps, so travis just uses superLU, thus when we convert the CSR matrix to int64, superLU complains.  Aside from maybe including umfpack as a dep, this also makes me think that we should test all of the possible scipy-solvers, like 'cg', 'gmres', etc just be sure that we catch any other errors (like the fact that cg returns a tuple).

jgostick:
The solution to this problem is to NEVER use superLU, so we'll add scikit-umfpack as a dependency (see #1004)

jgostick:
before merging this I'd like to add all the tests @magnaou has written for the iterative solvers.

PR #983, opened 2018-07-12
Fix bug in ReactiveTransport that set_source used to override Dirichlet BCs

ma-sadeghi:


jgostick:
@magnaou Can you take a look at the changes I made to ReactiveTransport and see if they should also be implemented in the transient classes?  Here is a summary of how I did it:

1.  When adding BCs, the `_set_BC` method of LinearTransport is called, so instead of messing with that class, which should be independent of reactions, I overloaded _set_BC on the ReactiveTransport class to first check that no sources are present, then call the super class to proceed with actually adding BCs.

2.  When adding sources, the `set_source` of ReactiveTransport is used, so I just added the check directly in there.

I guess my question is: Do the transient algorithms inherit from ReactiveTransport?  If so then they will receive these changes.  Or do you have custom versions of these steps?

PR #992, opened 2018-07-16
mhrz changes & bug fixes

magnaou:


magnaou:
@ma-sadeghi, to avoid having "0" tol as I'm using absolute values.

magnaou:
@ma-sadeghi I agree that using max dangerous...
I will think about it and update it.
Thanks

magnaou:
Again, as discussed with @ma-sadeghi , using "norm" or "max" will result in the same issue if the system of linear equs. has large magnitude gradients from one term to another. An alternative will be using "min", but this may result in slowing down the solver.
For now, I'm assuming that all terms have comparable orders of magnitude, and I'm using "norm".

PR #996, opened 2018-07-17
Cantera sandbox

jgostick:
This is a new PR for adding Cantera to calculate physical props by @ma-sadeghi


jgostick:
I'm demoting this from V2.1 to V2+ since they don't allow pip install yet

PR #1000, opened 2018-07-18
Unyt support

jgostick:
Adding tentative, proof-of-concept support for the unyt package.  Very few changes were actually required.  It only needed to not forcefully convert arrays to Numpy when writting, and tracking units when interleaving, and adding units when interpolating.  Not sure how far else down the rabbit hole this would need to go though.


jgostick:
They just made a JOSS paper for Unyt:

http://joss.theoj.org/papers/10.21105/joss.00809

jgostick:
We should also make this more general to handle other unit systems as well, such as Pint

PR #1021, opened 2018-07-26
Adding values for all properties, and converting constants to models

jgostick:


jgostick:
There is once 'gotcha' with this approach.  Now that all constants are being produced by models, if you lazily change a value like pn['pore.temperature'] = 444, when you regenerate models this would get set back to 298 (or whatever).

jgostick:
I have changed all the constant models back to plain constants.  Users of the GUI can add a constant model if they wish to change it.  This keeps actual constants behaving as such.

Issue #1041, opened 2018-07-31
from_neighbor_throats and from_neighbor_pores models have bugs


TomTranter:
[test_fnt_fnp.txt](https://github.com/PMEAL/OpenPNM/files/2245880/test_fnt_fnp.txt)

TomTranter:
These methods don't work for subdomain objects (geom and phys)

TomTranter:
@jgostick would it be worth adding methods to Base that find the parent and children of an object.
Currently project has find_phase and so on but you have to know what the object is to use the right one.
I'm thinking a generic method that works out what it is then calls the appropriate find method.

i.e phys.find_parent() returns phase
     geom.find_parent() returns net

If you ask a domain object to find its parent or a subdomain object to find its child then it can either return itself or maybe None. This is needed for the from_neighbor_pores method to be completely generic so I wonder instead of adding it inside that method to make it a more accessible one.

I've coded
def find_parent(obj):
    if 'Subdomain' not in obj._mro():
        # Network, Phase, Alg
        return obj
    else:
        prj = obj.project
        op_class = obj._isa()
        if op_class == 'geometry':
            # Geom
            return prj.network
        else:
            # Phys
            return prj.find_phase(obj)

TomTranter:
We decided to put the function in project as it is useful to have and actually Subdomain already has a method that is similar _get_boss which can be replaced.

TomTranter:
find_parent and find_child are valid from a graph theory point of view but it's new terminology for openpnm. Happy with it?

jgostick:
No, it conflicts with OOP!  Master/slave, professor/student (just kidding), boss/subordinate, full_domain/sub_domain, ???

PR #1044, opened 2018-08-01
Fix from_neighbor_pore and from_neighbor_throat misc models

TomTranter:
Addresses issue #1041


jgostick:
I am not sure that this `find_full_domain` was actually necessary.  Upon looking at your improved models, I see that your using `map_pores`.  In this case it does not matter whether you map to a network or a phase, since they're both 'full domain' objects, so you could just do `network.map_pores(pores, target)`.

TomTranter:
It does need it because if you use the models for a physics object it used to be trying to get data from the network instead of the phase

Issue #1040, opened 2018-07-30
Percolation algs still need work. LPF and curve points


TomTranter:
LPF is inconsistent across the algs - needs to be improved in MixedIP.

TomTranter:
@jgostick any particular reason partial_filling is on Porosimetry instead of OP. Also, I think the terminology is a bit misleading calling it partial filling. This to me sounds like partial saturation i.e starting invasion when some pores are already filled. What do you think about calling it fractional_occupancy which sounds more specific to a pore and throat rather than a network

TomTranter:
Also, there seems to be some confusion in Porosimetry. We have the ability to change 'late_pore_filling' in the setup method but then there's also a specific method set_partial_filling. Did you just forget to remove the late_pore_filling bit from setup. It seems like what's going on in set_partial_filling could be done in the setup

TomTranter:
Results is also totally different in MixedIP to other algs

TomTranter:
At the moment the results method for OrdinaryPercolation and InvasionPercolation just return occupancy and don't update the phase. OP takes a Pc and IP takes a Saturation as input (both could and should probably accept both). Porosimetry, on the other hand, does update the phase for the purposes of LPF with a pressure but not an occupancy and returns this like the previous two in a separate dict. MixedIP updates the phase with a bunch of info useful for post processing but doesn't return data in the same way or automatically update the occupancy with LPF like Porosimetry does. Therefore results can't be easily called in a loop to generate a curve and MixedIP has it's own method to do that without updating the phase. When an instrusion curve is generated using Porosimetry this calls results a bunch of times and will update the pressure on the phase if LPF is defined which isn't necessarily something we want to do.

TomTranter:
We need to apply LPF as a model to make it flexible but I think it should be applied to the algorithm, not the phase. This means we can run a simulation and calculate LPF intermittently without having to update the phase.

TomTranter:
Late pore filling should be a method

TomTranter:
[...]

PR #1045, opened 2018-08-01
Percolation housekeeping

TomTranter:
Address issue #1040


TomTranter:
@jgostick Lots of cosmetic changes to MixedIP to make it function a bit more like the other percolation algs for results, lpf and curve_plotting. I also have a method called copy_alg_data_to_phase as this is what results used to do but I'm happy to remove it and let users do this themselves if they every want to visualize the data in paraview. Maybe a few more little tweaks to do. I could also try to add back in coop pore filling but that requires more thought

Issue #1038, opened 2018-07-30
What to do about Boundary Pores


TomTranter:
I think we need to think about a proper treatment of boundary pores for both linear transport and percolation. And whether we need them and whether taking them out causes more problems than keeping them. Currently they are causing a lot of issues for the conductance models.

- They are useful for percolation when you want to start the simulation with zero saturation

- They are useful for transport for when you need a coplanar face for effective properties

- They are annoying as part of the model framework where they may not contain the data needed for some models to work and end up having unphysical values or fudged data

- They are easy to add to cubic networks but not random or extracted

So is it more useful and consistent to treat them as real pores in a sort of ultra high-porous material that sits directly in contact with the main domain and give them real diameters, volumes, conductances. Then when doing percolation and stuff it's simple enough to look at saturation in a bulk geometry rather than the whole domain. Or is it better to not give them any physical properties, maybe not even a geometry so that all the interleaved data returns nans and then give them special treatment in the conductance models - fudging the nans into real values when needed?

ma-sadeghi:
Here are some thoughts:
1. Would it make a difference to use boundary pores to calculate effective properties?
2. For percolation, would it hurt to start the simulation at some small-enough value, rather than 0?

TomTranter:
So for effective properties which is really treating the whole network as one large resistor or black box the effect of adding boundary pores is to say the edges of this box go through the centers of the boundary pores which affects the length and area in the calculation. So even though they shouldn't offer any extra resistance the calculation is going to return a different result compared to using the first set of real pores by virtue of having a slightly bigger box. The user can now specify the length and area in calc_eff_property so I guess they can decide where they want their box to be drawn. Another advantage to adding boundary pores is to make the box shape around an irregular internal structure (e.g. Voronoi). 

I would say starting percolation from zero is important but as I suggested above it can be handled using geometry labels. 

With the Voronoi, it's actually the throats at the surfaces of the shape that have real dimensions and the boundary pores connecting outside are purely fictitious. However, this is not always the case. I guess sometimes both could be real, like connecting to an external reservoir and sometimes both could be fictitious.

[...]

PR #1046, opened 2018-08-02
Start sorting out boundary pores

TomTranter:


TomTranter:
PR to address issues discussed in #1038

PR #1062, opened 2018-08-17
Restructuring conductance models

ma-sadeghi:


ma-sadeghi:
I'll post a summary of what I've done soon. Also, I'm writing a bunch of tests for the new methods that I added, will push them once they're ready.

TomTranter:
@ma-sadeghi, I see you've been working very hard on this, nice job. Here are my initial comments

- In all models I would use a parameter for pore_centroid rather than hard-coding pore.coords.
- throat_endpoints.cubic_pores breaks when there is more than one throat in the network:

```python
EP1 = xyz[cn[:, 0]] + 0.5 * D1 * unit_vec
```
should be

```python
EP1 = xyz[cn[:, 0]] + 0.5 * D1[:, _sp.newaxis] * unit_vec
```
This got through your tests because they only have one throat. I just pushed a fix but didn't update the tests which should be on a bigger network.

end_points should accept a throat_vector parameter instead of working out a unit vector from pore_coords, this should be a model on geoms. There is a simple throat vector model on my percolation_housekeeping branch for cubic networks as it's needed for a percolation model. As discussed Voronoi has different throat vectors from the pore to pore center and so would extracted nets.

I'll post this now but continue testing

TomTranter:
Couple more things:
- it seems unnecessary to have the circular_pores wrapper for spherical pores
- I've added another end-points model that takes a throat centroid, vector and length and works out the 
  end-points from that. It does no checks on pores. This is for the extracted nets and Voronoi
- I also made a small change to Boundary.
- The new end-points test has a couple methods for throat vector and centroid which will later become models when another PR I'm working on is merged. The idea of the test is to try to ensure that diffusion through a domain of fixed size is independent of boundary geom and number of pores and that properties are being consistently calculated

ma-sadeghi:
> throat_endpoints.cubic_pores breaks when there is more than one throat in the network

Nice catch.

> it seems unnecessary to have the circular_pores wrapper for spherical pores

It's for the 2d support that's going to be addd later.

> end_points should accept a throat_vector parameter instead of working out a unit vector from pore_coords, this should be a model on geoms.

I think it might be better that endpoints always assume throats are aligned with pore-to-pore vectors. Non-aligned throats only affect throat length which are automatically handled if `throat.centroid` is provided. It also helps for better visualization (see below: top is using throat normal vector, bottom is assuming pore-to-pore):
image

ma-sadeghi:
Two questions though: is throat length a computed property in Voronoi/extracted networks? Or is it just throat diameter? @TomTranter @jgostick

TomTranter:
Assuming throats are aligned with pore-to-pore vectors is wrong. I don't know what visualization you're talking about. What's wrong with having a model that replaces the assumption. You get the same result back for cubic networks

ma-sadeghi:
I meant Jeff's throat visualization directly in VTK.

> What's wrong with having a model that replaces the assumption.

Just to make things simpler, if it only accounts for the added throat length.

TomTranter:
The throat length in Voronoi is assumed to be the fiber diameter, but the vector is definitely not the pore to pore vector and the pore center to the throat center to pore center length is most certainly not the pore to pore distance. Throat vector is quite fundamental in Voronoi. For extracted it would be calculated by working out the endpoints but still doesn't have to line up with the pore to pore vector and you could actually get the scenario you drew at the top (if the throat were a little longer and connected to the pores)

ma-sadeghi:
One more thing, the whole point of adding throat endpoints was to calculate throat length. But, `throat.endpoints.straight_throat` uses throat length as input, which makes it circular.

ma-sadeghi:
> Throat vector is quite fundamental in Voronoi.

I don't disagree. I'm just saying that we introduced throat endpoints to make calculating throat length/conduit lengths more straightforward. So in that sense, throat vector is not fundamental in these calculations.

TomTranter:
The point of the end_points was so that conductance_lengths would have something concrete to work from in every scenario. We don't have one unified method of determining where a pore ends and throat starts. With Voronoi you have a 2d throat and the plane has a normal vector and you want to work out how far you should extend it for conductance calcs. With sticks and balls I guess the logic is you start with the pores as the definitive geometry object and work out where the throat fits in between. With extracted you have some continuous space and need the end points as the dividing line that you could calculate a whole bunch of different ways. So I disagree, it's not circular.

ma-sadeghi:
@TomTranter and I discussed this issue, and here's the summary:
1. For Voronoi fibers, it's essential that the throat endpoints be calculated according to the current `throat_endpoints.straight_throats` model.
2. `throat.centroid` needs to be added as an optional argument to the `throat_endpoints.spherical_pores` model to make it more general, in case throat centroid and pore coordinates are not co-linear.
3. `throat_endpoints.cubic_pores` remain unchanged as its use should be reserved for cubic networks only, rather than random networks, so `throat.centroid` and pore coordinates are always colinear.

ma-sadeghi:
By the way, here's a sketch of how the new models work:
image

PR #1063, opened 2018-08-18
Mstr v2 mhrz

jgostick:


magnaou:
I changed the base of the PR to "dev" instead of "master".
I also removed the new BCs stuff and put it in a new branch "newBCs". It was causing lots of issues...

PR #1068, opened 2018-08-21
Fix 1066 prop naming

jgostick:
Adding a check to setitem in Base so that trying to add 'pore.foo' will raise an exception when 'pore.foo.bar' is already present, and vice-versa.  (This time PRing to dev instad of master)


jgostick:
We still need to decide if the check should span objects.  For instance should `pn['pore.area.test']=1` be allowed if `geo['pore.area']` exists?  This doesn't seem to break anything, but does seem fragile.

jgostick:
this can wait and requires a bit more thought and discussion

jgostick:
I feel like the above should *not* be allows, just avoid confusion.  This PR should be updated to check across objects.

PR #1076, opened 2018-08-22
BCs

magnaou:
Impose Dirichlet BCs preserving "A" matrix symmetry.


magnaou:
@jgostick I think you can merge it. @ma-sadeghi has implemented a faster version!

PR #1087, opened 2018-08-24
Direction Dependent Quantities

TomTranter:
Add touch pressure to all meniscus models and handle in percolation algorithms


TomTranter:
To do:
- Add touch pressure to Purcell
- Handle directional dependent Pc in MixedIP
- Handle in Invasion Percolation (impossible in OP)
- Possibly think about how to handle more generally - tied to throat conns (already sort of doing this with conduit_length.head an .tail)

PR #1090, opened 2018-08-27
Add PyAMG solver + slight change in solver logic

ma-sadeghi:
Following the discussion with @jgostick , here's a summary of what I've done:

`alg.settings['solver']` used to be a string, either signifying one of `scipy`'s solvers (ex `cg`, etc.), or other solver _packages_, ex. `pyamg`, `petsc`. Solver configs such as `max_iter`, etc were all stored under `alg.settings`. Now, `alg.settings` requires to be a `dict` with keywords 'family', `solver`, etc. `family` could be either `scipy`, `pyamg`, or `petsc`, so on so forth. However, for convenience, if the user passes a string such as `cg` or even `pyamg` as `alg.settings['solver']`, the solver turns it into a proper `dict`, and then acts on it.


ma-sadeghi:
There's an issue with scipy's iterative solvers:

- `tol` not sufficiently small: gives gibberish with return code 0 (converged)
- `tol` sufficiently small: fairly accurate (up to 5,6 significant figures), with return code > 0 (diverged). This is expected since while the numbers are fairly accurate, they don't meet the strict requirement of the given `tol`.

So basically, either you set the tolerance large enough for the scipy solver to converge, in which case you get gibberish, or set it very small, which returns a fairly accurate solution, but since the exit code indicates "not converged", `GenericTransport` throws an error.

Workaround: set `tol` small enough but don't throw error if exit code is not 0. This is potentially dangerous since you can never be sure if the solution is correct.

ma-sadeghi:
Another issue that needs attention: what should we use as the default solver? (right now it's scipy's `cg`)

I'd say a direct solver such as `spsolve` because of two reasons:
- It's a direct solver, hence 100% accurate (at least to machine precision).
- It's faster than iterative solvers up until network size of ~100,000.

ma-sadeghi:
Btw, the issue with scipy's iterative solvers doesn't always happen, here's a code snippet that reproduces it:

```
import openpnm as op
import scipy as sp
import numpy as np
np.random.seed(3)
ws = op.Workspace()
ws.settings['loglevel'] = 40
proj = ws.new_project()

pn = op.network.Cubic(shape=[10, 10, 10], spacing=1e-4, project=proj)
geo = op.geometry.StickAndBall(network=pn, pores=pn.Ps, throats=pn.Ts)
air = op.phases.Air(network=pn, name='air')
water = op.phases.Water(network=pn, name='h2o')
hg = op.phases.Mercury(network=pn, name='hg')
phys_air = op.physics.Standard(network=pn, phase=air, geometry=geo)
phys_water = op.physics.Standard(network=pn, phase=water, geometry=geo)
phys_hg = op.physics.Standard(network=pn, phase=hg, geometry=geo)

perm = op.algorithms.StokesFlow(network=pn)
perm.setup(phase=water)
perm.set_value_BC(pores=pn.pores('left'), values=100)
perm.set_value_BC(pores=pn.pores('right'), values=0)
perm.run()
```

ma-sadeghi:
TODO: See why the SubclassedTransportTests changed value when switching from `cg` to `spsolve`.

ma-sadeghi:
This branch has already been merged onto the outflow BC branch, so I'm closing it.

PR #1103, opened 2018-09-10
Transient

magnaou:
- Set zero ICs as default **(done)**
- Rename the transient fields to something like: 'pore.quantity.t.0', 'pore.quantity.t.1e-1', etc **(done)**
- The algorithm accepts a scalar or a list to export data. When a scalar is given, it's considered as the export time step. When a list is passed, times in the list are used to export transient data **(done)**
- Export the transient data in a format easily exploitable by Paraview to play animations **(in progress)**

### DIFF
--- a/openpnm/algorithms/GenericTransport.py
+++ b/openpnm/algorithms/GenericTransport.py
@@ -534,6 +534,16 @@ class GenericTransport(GenericAlgorithm):
         d = {quantity: self[quantity]}
         return d
 
+    def results_transient(self):
+        r"""
+        Fetches the calculated quantity from the algorithm and returns it as
+        an array.
+        """
+        quantity = self.settings['quantity']
+        q = [k for k in list(self.keys()) if quantity in k]
+        d = {k: self[k] for k in q}
+        return d
+
     def rate(self, pores=[], throats=[], mode='group'):
         r"""
         Calculates the net rate of material moving into a given set of pores or
--- a/openpnm/algorithms/GenericTransport.py
+++ b/openpnm/algorithms/GenericTransport.py
@@ -5,6 +5,7 @@ import scipy.sparse as sprs
 import scipy.sparse.csgraph as spgr
 from scipy.spatial import ConvexHull
 from scipy.spatial import cKDTree
+from decimal import Decimal as dc
 from openpnm.topotools import iscoplanar
 from openpnm.algorithms import GenericAlgorithm
 from openpnm.utils import logging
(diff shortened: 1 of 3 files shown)


jgostick:
I have a better idea about naming transient data...using the @ symbol!  As in: `'pore.concentration@1e-5'`

jgostick:
Then the vtk output function can look for @ symbols in the dict name and assume they are transient time steps.  We can even enhance the __setitem__ subclass to forbid the use of @ in dict names, which we can circumvent using the `update` method (which is not subclassed at all to allow us to do things like this).

Also, the `results` method of the algorithm should take an argument indicate which time to return, so:

`alg.results(times=1e-4)` should return a single array, perhaps with the time stamp stripped off (to make it easier for accessing from the phase), and `alg.results(times=range(0, 1, 1e-3))` should return an array of all the results at the given times.  For flexibility `alg.results(steps=range(0, 100, 10))` could also be accepted, and return the requested steps rather than specific times so the user doesn't have to figure the exact times they want.

magnaou:
1) The use of `'@'` instead of `'.t.'` could be more elegant especially if we reserve the `'@'` symbol only for transient as you said.
2) For returning results, I agree, we should give the option to only return the time steps we need. I have implemented something that does return everything. So, it does something like: `alg.results(times=all)`.
3) Doing something like `alg['pore.conentration@5:1e-3']` means that you store both the step and the time in the dict key? isn't this redundant? do we need to store the step for something specific?

jgostick:
The step number is redundant, but could be helpful...for instance, you could figure out what the time step was since @5:13e-5 and @10:18e-5 means there were 1e-5 seconds per step.  Not sure if this is helpful or confusing.  I'm ok to just go with @13e-5.

magnaou:
I see. I'm ok with keeping it for now.

jgostick:
No let's get rid of it, it's just confusing.  The info after the @ should be purely time..I was just throwing around idea.  Let me know once you've got this and I'll adjust the VTK accordingly.

magnaou:
Just a comment, when we do `alg.results(times=range(0, 1, 1e-3))`,
1) `1e-3` should be multiple of `alg.settings['t_output']` if `alg.settings['t_output']` is a scalar, or,
2) `range(0, 1, 1e-3)` should results in values that `alg.settings['t_output']` includes if `alg.settings['t_output']` is a list.

magnaou:
Also, should we force the user to use the same format for time steps (i.e., scientific without ".")?
If yes, `alg.results(times=1e-1)`, will return the results but `alg.results(times=0.1)` will not.

magnaou:
Never mind, we have to accept all formats and convert them.

magnaou:
I'm done with these changes.

PR #1105, opened 2018-09-13
Improve connect pores to be able to connect multiple batches of pores in a 1 to 1 fashion

ma-sadeghi:
This PR is similar to what was recently done to `merge_pores`. Previously, `connect_pores` would create new connections between two lists of pores, `pores1` and `pores2`. This PR preserves the old functionality but adds the possibility that user sends in a list of lists of pores as `pores1` and `pores2`, and consecutively make connections between the corresponding batches in a 1 to 1 fashion. 

Example:
```
ps1 = [[1, 2], [3, 4, 5]]
ps2 = [[55], [9, 99]]
op.topotools.connect_pores(net, pores1=ps1, pores2=ps2)
>>> net['throat.conns'][-8::]
array([[ 1, 55],
       [ 2, 55],
       [ 3,  9],
       [ 3, 99],
       [ 4,  9],
       [ 4, 99],
       [ 5,  9],
       [ 5, 99]])
```
The reason behind this PR is similar to that for `merge_pores`. In `connect_pores`, there are multiple calls to `extend` which slows it down significantly. This PR potentially helps `merge_pores` as well since it includes multiple calls to `connect_pores`, which can now be done in a single call. Fixing `merge_pores` for further performance gain is WIP.

### DIFF
--- a/openpnm/topotools/topotools.py
+++ b/openpnm/topotools/topotools.py
@@ -1426,11 +1426,23 @@ def connect_pores(network, pores1, pores2, labels=[], add_conns=True):
            [32, 68]])
 
     '''
-    size1 = sp.size(pores1)
-    size2 = sp.size(pores2)
-    array1 = sp.repeat(pores1, size2)
-    array2 = sp.tile(pores2, size1)
-    conns = sp.vstack([array1, array2]).T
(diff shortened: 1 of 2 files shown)


jgostick:
merged onto #1130

PR #1120, opened 2018-10-10
Add linear_scale parameter to VoronoiFibers

TomTranter:


jgostick:
* Try fudging shape first, then scaling afterwards.  
* Also, I think the default should be no scaling.
* Think about making dt on each pore optional?

TomTranter:
@jgostick this contains the addition to VoronoiFibers and the fix to the labelling

PR #1114, opened 2018-10-09
Fixing issues sept11

jgostick:

PR #1130, opened 2018-10-16
Further improvements to merge_pores and connect_pores

ma-sadeghi:
`merge_pores` and `connect_pores` now support batch mode, i.e., you can pass a list of list of pores. The reason for doing this is because if you want to consecutively merge/connect a bunch of pores, the adjacency matrix is being rebuilt between each call to `merge_pores` and `connect_pores`. The old API still works, and here's how the new API works:

## Merge pores in batch mode
```
import openpnm as op
network = op.network.Cubic(shape[5, 5, 5])
pores = [[0, 1, 2, 5], [5, 10, 2, 11, 23]]
op.topotools.merge_pores(network=net, pores=pores)
```
After running the above snippet, both the batch of pores `[0, 1, 2, 5]` and `[5, 10, 2, 11, 23]` are merged into two new pores. Note that these two batches share some pores, and ideally the two new pores should be connected because of that. The improved `merge_pores` accounts for this.

## Connect pores in batch mode
```
import openpnm as op
network = op.network.Cubic(shape[5, 5, 5])
ps1 = [[0, 1], [2, 3]]
ps2 = [[5], [7, 9]]
op.topotools.connect_pores(pores1=ps1, pores2=ps2)
```
Running the above snippet results in the following new connections:
```
[0 , 5]     [2 , 7]    [3 , 7]
[1 , 5]     [2 , 9]    [3 , 9]
```

Finally, for the batch mode to work properly, the input pores must be a list of lists or a list of ndarrays, etc. The method throws an exception if input is not properly formatted.


jgostick:
This build is aborting because of the scikit-umfpack problem...you should merge the dev branch into this once I merge by PR #1114

PR #1140, opened 2018-10-25
Adding MercuryPorosimetry pre-made algorithm

jgostick:
The idea is to call this algorithm with only a network and geometry.  Not need to make Phases or Physics...or to specify inlets.  It's all done for you internally.  Takes 10 lines and turns it into 1.  This will also be a template for how to do other pre-set algorithms like RelativePermeability.....or even just AbsolutePermeability.


jgostick:
OR...add a wrapper/helper function to IO called im_from_openpnm?

jgostick:
I have merged this file into the ``add_metrics_module`` branch, now part of PR #1303 .

PR #1158, opened 2018-12-12
Fix Voronoi Hulls

TomTranter:
Catch the voxels that escape the hull image procedure and assign them to the nearest pore (using max_filter). Scikit-image flood fill would potentially be faster but is not implemented yet and speed is not so bad

### DIFF
--- a/openpnm/materials/VoronoiFibers.py
+++ b/openpnm/materials/VoronoiFibers.py
@@ -538,6 +538,10 @@ class DelaunayGeometry(GenericGeometry):
             verts = np.asarray(unique_list(np.around(verts, 6)))
             verts /= self.network.resolution
             self.inhull(verts, pore)
+        # Catch the voxels that escaped the hulls
+        max_h = ndimage.maximum_filter(self._hull_image, size=2)
+        mask = self._hull_image == -1
+        self._hull_image[mask] = max_h[mask]
         self._process_pore_voxels()
 
     def _process_pore_voxels(self):


TomTranter:
voronoi_porosity

TomTranter:
There is still an issue which is not simple to resolve but not a major problem. The VoronoiFibers class builds two images. One binary for the fibers and pore space and one like snow.regions for the pore volumes. There are also functions for calculating the domain volume and dimensions etc using the vertices which are aligned to the bounding box defined by the shape parameter used to instantiate the class. The vertices are used to fill the image in certain locations and then a bresenham line function is used to create the fiber skeleton which is then dilated. However the shape parameter of the domain can translate to number of pixels in the image in different ways. The vertices are considered to be at cell centers in which case for a cube domain with length 2 and pixel resolution 0.1 you need 21 pixels in each dimension. Now when you sum the size of the image you get 21^3 / 0.1^3 which is not 2^3. The pore volumes are therefore slightly larger than their geometric equivalents using the polygon volumes defined by the vertices. If the resolution is small the problem isn't too great. I don't really know how to address it either. If the pixel edges are used instead to define the volume then translating a vertex coordinate to a pixel is trickier

TomTranter:
```
import openpnm as op
import numpy as np
import porespy as ps
from scipy import ndimage as ndi

np.random.seed(1)
prj = op.materials.VoronoiFibers(num_points=5,
                                 fiber_rad=0.2,
                                 resolution=0.01,
                                 shape=[2, 2, 2],
                                 name='test3',
                                 linear_scale=[1, 1, 2])
net = prj.network
del_geom = prj.geometries()['test3_del']
B1 = net.pores(['top', 'delaunay'], mode='xnor')
B2 = net.pores(['bottom', 'delaunay'], mode='xnor')
assert del_geom.vertex_dimension(B1, B2, 'volume') == 8.0
assert del_geom.vertex_dimension(B1, B2, 'length') == 2.0

h = del_geom._hull_image
f = del_geom._fiber_image
hf = h.copy()
hf[f==0]=-2

vox_poro = np.sum(f)/np.size(f)
hull_poro = np.sum(hf > -1)/np.size(hf)
pore_poro = np.sum(del_geom['pore.volume']/8.0)
print(vox_poro, hull_poro, pore_poro)
```

TomTranter:
@jgostick thoughts? @ma-sadeghi was this your problem?

PR #1174, opened 2018-12-25
nudging endpoints to prevent 0 length throats

jgostick:


jgostick:
This is an extremely minor, but critical change, to ensure that throat lengths are not zero in the endpoints models.  The snow algorithm actually calls these models (via import from openpnm), so if we want to fix the snow bugs, we need to merge and deploy this asap.

PR #1182, opened 2018-12-27
fixing performance regression in Cubic

jgostick:


jgostick:
I fixed the speed problem and also took the opportunity to improve the way spacing and shape are found on cubic networks.  I think its now 'fully' robust, meaning that it will work on networks with uneven spacing in each direction, networks that have been rotated, and even skewed....although the last two I haven't actually tested yet.

jgostick:
I think that is a pretty good idea, but...it breaks backward compatibility.  What about finding surface throats mathematically instead of using the adjacency matrix?

PR #1199, opened 2019-01-16
Statoil update

TomTranter:


TomTranter:
These changes get us close to the answers given by the poreflow code. @Zohaib-Atiq is working on compiling their code so we can debug it to find all the differences but it's challenging as it relies on some old libraries.

PR #1204, opened 2019-01-20
Return sub dicts

jgostick:
Since we allow subdicts to be written, we should probably support reading them too.  This PR allows the following:

```
pn['pore.foo.bar'] = 1
pn['pore.foo.baz'] = 2
d = pn['pore.foo']
d.keys()  # will be ['pore.foo.bar', 'pore.foo.baz']
```

This works with all the interleaving and multiple subdomains


jgostick:
There are 2 questions that need to be discussed:
1.  At present this returns a dict with FULL names, rather than just the nested dict names.  For instance:
```
pn['pore.foo.bar'] = 1
pn['pore.foo.baz'] = 2
d = pn['pore.foo']
print(d.keys()) 
['pore.foo.bar', 'pore.foo.baz']
```
This *could* contain just 'baz' and 'bar'.  

2.  Also, at present the look-up of nested dict names just uses the startswith attribute. For instance:
```
d = pn['pore.foo.ba']
print(d.keys())
['pore.foo.bar', 'pore.foo.baz']
```
In other words, the lookup of incomplete dict keys does not use the '.' as a delimiter, but just looks for matching strings.  This is sort of a bug

PR #1212, opened 2019-02-01
Add PerGeos IO class

jgostick:
This adds functionality for loading AND saving between openpnm and avizo, which uses the PerGeos format.


neumannrf:
I noticed I could only import the new IO class _after_ installing `unyt` and `flatdict` on my own, since it does not feature in any of the `*_requirements.txt` files. Shouldn't it be added to `test_requirements.txt`?

They **do**, however, appear in `setup.py`. So their absence does not affect users installing the package, but instead developers creating a environment from the requirement files.

PR #1220, opened 2019-03-15
Optimize Cubic network generation speed

ma-sadeghi:
This PR optimizes Cubic network generation speed. The optimization was done by replacing the list comprehension for generating pore coordinates by a numpy-vectorized version. This small modification makes Cubic network generation about 2.5x faster. Here's a speed comparison between the two branches (`dev` and `optimize_Cubic`) for different network sizes ranging from 10 thousand to 10 million pores.
performance

### DIFF
--- a/openpnm/network/Cubic.py
+++ b/openpnm/network/Cubic.py
@@ -11,6 +11,7 @@ from openpnm.network import GenericNetwork
 from openpnm import topotools
 from openpnm.utils import logging
 logger = logging.getLogger(__name__)
+from openpnm.utils.misc import tic, toc
 
 
 class Cubic(GenericNetwork):
@@ -100,8 +101,10 @@ class Cubic(GenericNetwork):
             spacing = sp.concatenate((spacing, [1]))
         self._spacing = sp.ones(3)*sp.array(spacing, ndmin=1)
 
-        points = np.array([i for i, v in np.ndenumerate(arr)], dtype=float)
-        points += 0.5
+        points = np.vstack(np.meshgrid(np.arange(shape[0]),
+                                       np.arange(shape[1]),
+                                       np.arange(shape[2]))).reshape(3,-1).T
+        points = points.astype(float) + 0.5
 
         I = np.arange(arr.size).reshape(arr.shape)
 
@@ -141,9 +144,11 @@ class Cubic(GenericNetwork):
 
         I = np.arange(arr.size).reshape(arr.shape)
         tails, heads = [], []
+        # tic()
         for T, H in joints:
             tails.extend(T.flat)
             heads.extend(H.flat)
+        # toc()
         pairs = np.vstack([tails, heads]).T
 
         super().__init__(Np=points.shape[0], Nt=pairs.shape[0], name=name,
--- a/openpnm/network/Cubic.py
+++ b/openpnm/network/Cubic.py
@@ -100,10 +100,10 @@ class Cubic(GenericNetwork):
             spacing = sp.concatenate((spacing, [1]))
         self._spacing = sp.ones(3)*sp.array(spacing, ndmin=1)
 
-        points = np.vstack(np.meshgrid(np.arange(shape[0]),
-                                       np.arange(shape[1]),
-                                       np.arange(shape[2]))).reshape(3,-1).T
-        points = points.astype(float) + 0.5
+        z = np.tile(np.arange(shape[2]), shape[0]*shape[1])
+        y = np.tile(np.repeat(np.arange(shape[1]), shape[2]), shape[0])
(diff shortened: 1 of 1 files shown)


jgostick:
Can you confirm that all the different 'connectivity' options still work?

ma-sadeghi:
To check if it works for multiple cases, I just added the following snippet in `Cubic` class init file:
```
flag_validation = sp.allclose(points, points2)
print("Shape:", shape, "Connectivity:", connectivity, flag_validation)
if not flag_validation:
    raise Exception("The two approaches didn't match!")
```
and ran the following script:
```
num_tests = 1
for i in range(num_tests):
    shape = np.random.randint(low=5, high=20, size=3)
    for conn in [6, 8, 12, 14, 18, 20, 26]:
        net = op.network.Cubic(shape, connectivity=conn)
```
which generates `N` Cubic networks with randomly assigned shapes, and for each network tests all possible connectivity values, and it passed:
```
Shape: [16 10 13] Connectivity: 6 True
Shape: [16 10 13] Connectivity: 8 True
Shape: [16 10 13] Connectivity: 12 True
Shape: [16 10 13] Connectivity: 14 True
Shape: [16 10 13] Connectivity: 18 True
Shape: [16 10 13] Connectivity: 20 True
Shape: [16 10 13] Connectivity: 26 True
```
To double check, I also ran a stress test with `N = 100`, and that also passed. So, I think the two approaches give similar results for a variety of combinations of network shapes and connectivity values.

ma-sadeghi:
@jgostick Actually, I just realized that above tests weren't needed. The vectorization is only for generating pore coordinates, which is independent of the connectivity and only depends on shape. So, I think it should be safe.

jgostick:
So this is only for making coordinates?  What about this:

```
Nx = 10
Ny = 20
Nz = 30
i = np.arange(0, Nx*Ny*Nz)
X = np.remainder(i, Nx)
Y = np.remainder(np.int_(i/Nx), Ny)
Z = np.remainder(np.int_(i/(Nx*Ny)), Ny)
```

ma-sadeghi:
> So this is only for making coordinates? What about this:
> 
> ```
> Nx = 10
> Ny = 20
> Nz = 30
> i = np.arange(0, Nx*Ny*Nz)
> X = np.remainder(i, Nx)
> Y = np.remainder(np.int_(i/Nx), Ny)
> Z = np.remainder(np.int_(i/(Nx*Ny)), Ny)
> ```

Two things: it's not as quick, and the order of coordinates doesn't match, although the latter is not really important since it returns "all" of the points, but in a different order (also I think there's a typo in your last line where you have `..., Ny)`, I think it should have been `..., Nz)`). Anyways, here's how the order of coordinates differ in the two approaches:
```
Nx = 2
Ny = 3
Nz = 4
shape = [Nx, Ny, Nz]

tic()
i = np.arange(0, Nx*Ny*Nz)
X = np.remainder(i, Nx)
Y = np.remainder(np.int_(i/Nx), Ny)
Z = np.remainder(np.int_(i/(Nx*Ny)), Nz)
P1 = np.vstack([X, Y, Z]).T
toc()

tic()
z = np.tile(np.arange(shape[2]), shape[0]*shape[1])
y = np.tile(np.repeat(np.arange(shape[1]), shape[2]), shape[0])
x = np.repeat(np.arange(shape[0]), shape[1]*shape[2])
P2 = np.vstack([x, y, z]).T
toc()

print(np.allclose(P1, P2))
```
Output (the second half to the right is how the order used to be, and the first half is generated by your snippet):
```
False
[[0 0 0 0 0 0]
 [1 0 0 0 0 1]
 [0 1 0 0 0 2]
 [1 1 0 0 0 3]
 [0 2 0 0 1 0]
 [1 2 0 0 1 1]
 [0 0 1 0 1 2]
 [1 0 1 0 1 3]
 [0 1 1 0 2 0]
 [1 1 1 0 2 1]
 [0 2 1 0 2 2]
 [1 2 1 0 2 3]
 [0 0 2 1 0 0]
 [1 0 2 1 0 1]
 [0 1 2 1 0 2]
 [1 1 2 1 0 3]
 [0 2 2 1 1 0]
 [1 2 2 1 1 1]
 [0 0 3 1 1 2]
 [1 0 3 1 1 3]
 [0 1 3 1 2 0]
 [1 1 3 1 2 1]
 [0 2 3 1 2 2]
 [1 2 3 1 2 3]]
```
And here's how they compare performance-wise, for a 200-cubed network:
```
Elapsed time in seconds:  0.4571712017059326
Elapsed time in seconds:  0.14392352104187012
```

jgostick:
ok, then this is all good.  I will merge.

PR #1222, opened 2019-03-28
new pore scale model

Zohaib-Atiq:
new conical and frustum model


jgostick:
This PR probably has too many conflicts to make it worth the trouble to merge now.  I suggest taking your function and putting it into a new branch, based on the latest version of dev, then deleting this branch.  I'm going to close the PR, so you can open a new one when the function is fully ready.   Don't forget to add tests!

Issue #1126, opened 2018-10-15
scaling on plot_connections is messed up in 2D


jgostick:
When plotting coordinates in 2D, it scale from 0,0 but when plotting connections, it zooms in to 0.5, 0.5

Issue #623, opened 2016-08-08
``plot_coordinates`` always returns a 3d plot even when passed a 2d network.


ma-sadeghi:
``plot_coordinates`` seems to always return a 3d plot even when passed a 2d network. It would be ideal for it to return a consistent plot based on the shape of the passed network.

jgostick:
All networks in OpenPNM are 3D.  Even a single layer of pores still has a z-coordinate.  

Perhaps there is a way to tell MatPlotLib to view the graph in 2d, like fig.view(blah, blah)?

ma-sadeghi:
I know, but still it gives the user a better experience if the method plots consistently.

jgostick:
Try to find the command that controls the view angle or camera angle of the
3d axes, and we can either add it to the documentation of the method, find
a way to invoke it automatically if the input points are 2D.  I think it
should still be a 3d plot, but just viewed in 2d, no?  I think that's how
Matlab works.  If you grab any 2d scatter plot and drag it, you get a 3d
view.

ma-sadeghi:
Why aren't we using `networkx` for plotting? It's fairly straightforward and customizable.

jgostick:
1. It doesn't let you chose the locations of the pores/nodes.  It puts them
wherever they make the least messy graph.
2. It doesn't support 3d graphs, they all lie in a plan.  Same problem with
Gephi and other tools.

Besides, the matplotlib solution makes the most sense for us since we
already have it as a dependency, and it's the de facto python math plotting
package so people generally know how to use it.

jgostick:
@ma-sadeghi , do you think your networkX graphs should be added to the code somehow?  Are they faster than the matplotlib approach?  Are they prettier?  Do they interleave with the existing plot tools, or will they be new methods?  Are they just meant for 2D?  Do they offer better rotation behavior in 3D?

jgostick:
I"m going to close this is since it's really just about setting the view angle on the Matplotlib window.  The use can drag the graph to the view they like.  It's also possible to manipulate the matplotlib axes as needed.  Here is a receipe (I will put this into the documetation for the methods):

``` python
import OpenPNM as op
import matplotlib.pyplot as plt
net = op.Network.Cubic(shape=[5, 5, 1])
fig = op.Network.tools.plot_connections(network=net)
ax = fig.axes()[0]  # Get handle to the subaxis containing the plot
ax.elev = 90  # Set the camera elevation
[...]

PR #1235, opened 2019-04-09
Refactor plot_conns and plot_coords, fix scaling issue plot_conns in 2D

ma-sadeghi:
This PR addresses issues #1126 and #623. The scaling problem in 2D for `plot_connections` is now fixed. Also, if 1/2D networks are passed, no longer the plot is 3D. I also refactored `plot_coordinates` and `plot_connections` a little bit.

### DIFF
--- a/example_PNP.py
+++ b/example_PNP.py
@@ -1,3 +1,4 @@
+"""
 import openpnm as op
 import numpy as np
 ws = op.Workspace()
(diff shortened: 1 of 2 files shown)


jgostick:
The changes look good, and nicely condensed.  My only question is how you make 3D plots without importing that Axes3D module from matplotlib....did they finally fix that terrible approach in V2?

ma-sadeghi:
@jgostick, Yes I think they have fixed it. I double checked and I confirm that I can make 3D plots without importing that module.

PR #1242, opened 2019-04-16
Support for transient data

magnaou:
Export transient data in XDMF format.
1- Supports transient and steady-state data
2- Steady-state data (such as geometrical properties) are not duplicated over transient files.
3- On Paraview, the Steady-state data are present at all time steps by pointing the "xmf" file to the initial time step file "hdf" which contains all the steady-state data.
4- Removed the option to export VTK files with time series as it doesn't properly handle non-uniform time steps AND generate large files...
5- When times series are present in phases, proj.export_data method will ALWAYS export in XDMF format an print a warning if the user requested to export into another format.

### DIFF
--- a/openpnm/utils/Project.py
+++ b/openpnm/utils/Project.py
@@ -5,8 +5,10 @@ import numpy as np
 import openpnm
 from copy import deepcopy
 from pathlib import Path
-from openpnm.utils import SettingsDict, HealthDict, PrintableList, Workspace
+from openpnm.utils import (SettingsDict, HealthDict, PrintableList, Workspace,
+                           logging)
 ws = Workspace()
+logger = logging.getLogger(__name__)
 
 
 class Project(list):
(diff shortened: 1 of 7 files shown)


jgostick:
@magnaou did you want to remove transient support from the VTK writer before I merge this?  Or should we handle that in a separate PR?

openpnm/utils/Project.py
```diff
@@ -579,28 +581,40 @@ def export_data(self, phases=[], filename=None, filetype='vtp'):
         network = self.network
         if filename is None:
             filename = project.name + '_' + time.strftime('%Y%b%d_%H%M%p')
-        # Infer filetype from extension on file name...if given
+        # Check if any of the phases has time series
```
jgostick: I don't like checking this here.  I think this should be checked within each specific io class, or better yet, should be added to the GenericIO class so that all classes get it.  For instance, add a method called ``is_transient`` and they add a check to all classes that raise an exception if it returns ``True``.  It's a bit more work, but more general and maintainable going forward.  Also, the ``phases`` argument should accept a single object or a list of objects, so be sure to check for that.

magnaou: @jgostick Thanks for reviewing. I agree, it's better.
You want to raise an exception if transient data are present and the user wants to export in format other than XDMF? Isn't raising a warning and exporting in XDMF format better?

jgostick: no, raise an exception, with a message to use xdmf.  You're way is more convenient for sure, but it's not a good idea to write a bunch of files that the user didn't ask for.

magnaou:
@jgostick Yes, I removed transient support from VTK files.

PR #1275, opened 2019-05-09
just delete and replace the origninal code

Ni2M:
Estimation of K in two different rock samples used in Pore Scale Imaging and Modeling paper.


Ni2M:
@jgostick  An example of paper recreation part a : estimation of K for two rock samples is ready to merge

PR #1286, opened 2019-05-27
Upload files for hysteresis paper recreation

TomTranter:


TomTranter:
@jgostick These notebooks work and can be merged

Issue #1284, opened 2019-05-26
Add a "set_label" function to Base?


jgostick:
I've always found it a bit annoying that we had to do:

```
net['pore.label'] = False
net['pore.label'][Ps] = True
```

If we added a new method to Base, it could be like this:

```
net.set_label(label='label', pores=Ps)
```

This could (a) create 'pore.label' if not present, then add label to locations ``Ps``.  We could also make it to pores and throats simultaneously, and also add a mode argument so that it could be used to 'add', 'overwrite', or 'remove' labels.

jgostick:
this is addressed in PR #1289

PR #1289, opened 2019-05-30
Added set_label method to simplify manipulation of labels

jgostick:
This addresses is #1284


jgostick:
@ma-sadeghi can you take a look at this and make sure it does the things you'd expect from it? As it is now, you can add labels to pore and/or throats in 'overwrite' or 'merge' mode, or remove from pores and/or throats.  You can also purge the entire label from all pores and/or throats.

PR #1309, opened 2019-07-17
Metrics relperm

Ni2M:
metrics relperm:
user defines phase & runs IP
call metrics.relativepermeability
get the relperm curves in 3 directions (assume we have percolating cluster in all 3 dimensions)


Ni2M:
The code works well for K_r calculation in three dimensions.

jgostick:
@Ni2M Congrats on getting this PR working and passing all tests!

Ni2M:
Thank you! The code is ready to use. The tests are more related to the boundaries, because we assume the invasion percolation is performed by the user in this algorithm.

PR #1336, opened 2019-11-10
Optional physics

### LINKED ISSUE #1311: Make geometry argument optional in Physics initialization
Default should be ``None``, while will check is 1 geometry is present and use it, if no geometries are present it will use network, and if 2 or more geometries are found it will throw an error.

jgostick:
Closes #1311. As suggested in issue #1311 this PR changes the Physics initialization to optionally accept no Phase or Geometry argument.  

If phase is not given, then a GenericPhase is generated and associated with the Physics

If geometry is not given, then one of the following occurs:
(1) If no geometry exists in the project, then a GenericGeometry is created
(2) If one geometry already exists then that one is used
(3) If multiple geometries exist an error is thrown.

The PR also add ``set_phase`` and ``set_geometry`` methods to the GenericPhysics class.  These methods can be used to swap which phase and geometry objects a physics is associated with.


jgostick:
I'm closing this, since it's a bad idea and breaks stuff.  But I'm going to open a new issue with an alternative approach.

PR #1348, opened 2019-12-08
Add imported geometry example redo

jgostick:


jgostick:
Wow, it actually all works!  Not sure why travis is failing.  It seems like it's running the tests against the wrong version of openpnm.

jgostick:
I think I know the problem.  I added porespy as a testing dependency, since it's used to do some network extractions.  I think during the installation, it must grab openpnm from pip, instead of using the actual repo its testing.  Not sure why github does not do it though.

jgostick:
This branch got way behind, so I've moved the example notebook onto the ``add_porespy_io`` PR, since they are closely related.

PR #1371, opened 2020-01-15
Hotfix: dispersion rate is not accurate

ma-sadeghi:
+ Overloading `GenericTransport.rate` for `AdvectionDiffusion` algorithm was unnecessary with our latest improvements on `GenericTransport.rate`.


jgostick:
Not sure about this one.  Ask @magnaou to review it.  He had to write a custom ``rate`` method for advection-diffusion at one point, so should be up to speed with the situation.

ma-sadeghi:
@jgostick: I think something's wrong with the overloaded rate method. I realized this when trying to debug Tony Cuenca's student: when the RFB model seemed to converged, `oc.rate(pores=source_pores)` exactly matched the net current calculated explicitly using the BV kinetics in source pores, whereas it was way off compared to `ad.rate(pores=source_pores)`. After I removed the overloaded rate method, the values perfectly matched. If you recall, the most recent change to `GenericTransport.rate` was to make it compatible with unsymmetrical A matrices (like dispersion for instance). Still, I'm not sure. At least I expected that the overloaded rate method to also work correctly.

@magnaou Can you please take a look at this?

ma-sadeghi:
@jgostick I looked into this issue further and I think everything is (was) alright, although found an actual _workflow_ bug (see issue #1372). The reason I was getting different values from ad.rate() and oc.rate() had nothing to do with the overloaded rate method. In fact, it was because after `StokesFlow.run()` is called, the `physics` object never gets regenerated, and so the `ad_dif_conductance` pore-scale model is using a uniform pressure field to calculate `throat.ad_dif_conductance`, hence the discrepancy between `ad.rate()` and `oc.rate()` even when the RFB simulation seemed to have converged! (see the recent issue I created #1372)

Anyways, I compared the values from the overloaded rate method with those from `GenericTransport.rate` for a 1D toy model, for which the analytical solution is known, and can now confirm that both give identical values. I also compared the two on Tony Cuenca's student's RFB model, and got identical rate values from both rate methods.

So, I think the overloaded rate method is redundant, although it's a relief that my initial guess that it gave **incorrect** rate values turned out to be wrong.

ma-sadeghi:
Btw, here's the code snippet of the toy model for reference
```
import numpy as np
import openpnm as op


ws = op.Workspace()
ws.settings["loglevel"] = 50
s_scheme = "hybrid"
d = 1.0e-6

net = op.network.Cubic(shape=[1, 10, 1], spacing=d)
geom = op.geometry.GenericGeometry(network=net)

geom["pore.diameter"] = d
geom["pore.area"] = np.pi * d**2 / 4
geom["throat.conduit_lengths.pore1"] = d/4
geom["throat.conduit_lengths.throat"] = d/2
geom["throat.conduit_lengths.pore2"] = d/4
geom["throat.diameter"] = d
geom["throat.area"] = np.pi * d**2 / 4

air = op.phases.Air(network=net)
phys_air = op.physics.Standard(network=net, geometry=geom, phase=air)
ad_dif = op.models.physics.ad_dif_conductance.ad_dif
phys_air.add_model(propname="throat.ad_dif_conductance", model=ad_dif,
                   s_scheme=s_scheme)

sf = op.algorithms.StokesFlow(network=net, phase=air)
sf.set_value_BC(pores=net.pores("left"), values=1e5)
sf.set_value_BC(pores=net.pores("right"), values=0)
sf.run()
air.update(sf.results())

phys_air.regenerate_models()

ad = op.algorithms.AdvectionDiffusion(network=net, phase=air)
ad.setup(s_scheme=s_scheme)
ad.set_value_BC(pores=net.pores("left"), values=1)
ad.set_value_BC(pores=net.pores("right"), values=0)

ad.run()
air.update(ad.results())

c = air["pore.concentration"]

gd = phys_air["throat.diffusive_conductance"].mean()
Q = sf.rate(pores=0)[0]
Pe = Q / gd
i = 5
mdot = Q * (c[i] + (c[i] - c[i+1]) / (np.exp(Pe) - 1))

mdot_internal = ad.rate(pores=5)[0]
mdot_inlet = ad.rate(pores=0)[0]
mdot_outlet = ad.rate(pores=9)[0]

print(f"RelErr in mass consv.: {abs(mdot_inlet + mdot_outlet) / mdot_inlet * 100:.4f} %")
print(f"RelErr in mdot (vs. exact sln): {abs(mdot - mdot_inlet) / mdot * 100:.4f} %")
print(f"Absolute value of mdot: {mdot_inlet:.5e}")
```

PR #1404, opened 2020-02-15
Improve docstrings in various locations, with help of docrep

jgostick:
docrep is a pretty great tool that is not as well known as it should be.  I have used it to inherit docstrings from parent methods, and then add new info to the child class docstrings.  It's not implemented everywhere yet, but slowly we can update docstrings throughout the package.


jgostick:
This PR in it's current form requires us to drop support for Python 3.6.  I'm using the `dataclasses` module from the standard library to define the settings.  It provides two very nice features:

1. You can add a simple class definition to a file, with settings and their default values (and types) defined as class attributes.  This would be an alternative to the dictionary we are currently using.  In this regard it does help much, with the main advantage being that we can define an empty list as a default setting without having to use a deepcopy each time we instantiate a new object.
2. Most importantly you can add a docstring to the class, which can then be added to various places for users to inspect.  It really makes the documentation much more coherent.

jgostick:
I abandoned the dataclasses approach, and am just using a normal class.  The use of a class is necessary for 2 reasons.  It gives us a place to write doc strings for the settings and it allows us to use docrep to inherit the docs of one settings class onto another.  It's all very tight now!

PR #1430, opened 2020-03-11
Add porespy to io module

jgostick:
A new PoreSpy class is added to the ``io`` module, which takes a pickled dictionary from PoreSpy, strips off all assumed properties, then passes the network to the Imported geometry class, which moves all props to itself and add models for anything that's missing.


jgostick:
This is ready, but could probably use some discussion to make sure I didn't miss anything

jgostick:
Make sphere/cylinder models adjustable via settings, move conductance models to own PR

PR #1441, opened 2020-03-13
Parallel computing with PETSc

magnaou:
fixed previous issues related to:
- compatibility with the new petsc version (3.12)
- availability of both multithreading and multiprocessing


jgostick:
I think writing tests for this wrapper is probably a good idea, since the petsc4py team has shown they don't mind pushing breaking changes

ma-sadeghi:
I tested the modified wrapper, and indeed it can be run without breaking on many-core clusters via
```
mpirun -n num_cores python openpnm_script.py
```
There's only one issue left, that is scaling. I tested it on Compute Canada using a simple diffusion script with ~3.5M pores, and compared it with a demo script which solves the diffusion equation with a nonlinear source term on a 3D lattice possibly using finite-difference, and here is the result:
petsc_scaling
I don't know what exactly is holding us back from a nice close-to-ideal scaling.

I'd say let's merge this PR, and open up another issue regarding the scaling issue.

ma-sadeghi:
For reference, here are the two scripts that I used for the scaling test: [bratu3d.py](https://gist.github.com/ma-sadeghi/b0b73e579d0836afd633860cc526107c), [pnm_petsc.py](https://gist.github.com/ma-sadeghi/45044acdda2658d7f67f68e13fd4d5c1)

magnaou:
@ma-sadeghi Thanks for the comparison. I have 2 comments:
1) In `bratu3d.py`, an initial guess is first computed and then passed to the `cg` `petsc` solver which I think we are not doing. This could explain why their solution is converging faster (`cg` requiring less iterations on `bratu3d.py` than `pnm_petsc.py`). We could do the same test with a direct solver such as `mumps` and see what happens. Another option would be forcing `bratu3d.py` to a `zero` initial guess. Anyways, if this is the reason, we should consider implementing a similar `initial_guess` method into `OpenPNM`.
2) In 'pnm_petsc.py', running multiple algorithms inside a `for` loop is really not efficient. I have talked with both of you (@jgostick @ma-sadeghi) about the memory not being cleared even after deleting the `project`, `network`, etc. In addition, I have noticed that when running multiple `algorithms` of **the same size** inside a `for` loop, the `s/iteration` increases with the iterations probably caused by memory issues where arrays can't be efficiently stored and need to be chunked...

ma-sadeghi:
@magnaou 
1. I tested `bratu3d.py` with zeros as initial guess, and it didn't change its scaling.
2. Right, but `nvar = [150]` only has one element, so it's not actually a `for` loop. Regarding the memory leak, you're right, I think there might be some loose ends, for which we can open up another issue.

magnaou:
Ok, I see. Could this scaling issue be caused by the overhead of other `openpnm` methods...
Do you think we if compare only the time taken by `_solve`, we should see a better scaling?

PR #1443, opened 2020-03-14
Renovate pnp algs

jgostick:
Changing the names of the PNP related algorithms to be more descriptive, including introducing the term Multiphysics to the PNP system solver.  Also changes the inheritance structure so the Multiphysics algs are instances of GenericAlgorithm, not any of the transport classes.


jgostick:
@magnaou and @ma-sadeghi : what do you think of this new naming?

```
ChargeConservation --> IonicConduction
IonicTransport --> NernstPlanckMultiphysics
```

And the transient versions are:

```
TransientChargeConservation --> TransientIonicConduction
TransientIonicTransport --> TransientNernstPlanckMultiphysics
```

Also, the ``NernstPlanckMultiphysics`` algs now descend from ``GenericAlgorithm`` instead of ``ReactiveTransport``.  So it now has far less unneeded methods like boundary conditions setters.

magnaou:
`ChargeConservation --> IonicConduction --> IonicTransport` since it's not really conduction?

jgostick:
Should we offer wrapper/helper classes to the old names, or just stop supporting them?

jgostick:
How is it not conduction?  I thought it was basically Ohm's law?

ma-sadeghi:
> Should we offer wrapper/helper classes to the old names, or just stop supporting them?

I'd go with stop supporting old names.

magnaou:
@jgostick The charge conservation equation reduces to Ohm's law in certain circumstances only (electroneutrality and absence of concentration gradients). In addition to these 2 conditions, if we consider a constant conductivity of the solution, the charge conservation equation reduces to a Laplace equation of the potential. From Newman's book, > When there are concentration gradients, the current density is not proportional to the electric field, and Ohm's law does not hold
I'm sorry about the confusion, I know we discussed the naming several times.

jgostick:
Yes, but the algorithm is basically Ohm's law with some source terms added via the ``NernstPlankMultiphysics`` system solver.  As far as I'm concerned this is still Ohms law.  Constant conductivity of the solution is also not a requirement of Ohms law.  Recall that the algorithm name refers to the physics we are applying to the pore-throat-pore conduit, so if you put a voltage field and some current moves, that is Ohms law.  Putting a source or sink term in the pores does not change things either...Ohms law applies when power sources are present in the system, which is essentially a source term.

magnaou:
Fair enough, let's stick to the initial plan then; `ChargeConservation --> IonicConduction`.

jgostick:
The ``IonicConduction`` algorithm on it's own applies voltage and current flows....Ohms law.  When this algorithm is coupled with diffusion and nernst-planckian effects then the statement you quoted from Newman is right.  But this algorithm itself is still only responsible for the voltage field part.  Correct?

magnaou:
Exactly. It solves for the potential field. So, as you said, it's Ohm's law when used alone.

Issue #1393, opened 2020-02-04
Better handling of iterative properties


ma-sadeghi:
At the moment, the user needs to manually specify which properties are iterative. This is fine since we can't/shouldn't make any assumptions in this regard. But, once a property is declared as `iterative_prop` by the user, `OpenPNM` **should** look up the dependency hierarchy and set upstream properties as `iterative_prop` as well. A good example is conductance, which could potentially depend on iterative properties.

jgostick:
Technically it's possible since we have the dependency tree now, so we could look up all the parent nodes of the tree and regenerate them  if needed.  

However, I don't quite understand the use case.  Consider:

   B
 /
A -- C -- D
 
A depends on B and C, C depends on D.

If A is iterative should we regenerate B, C and D?

If A is *not* iterative, but *B* is iterative, we should regenerate A and B, but what about C and D?

ma-sadeghi:
> If A is iterative should we regenerate B, C and D?

I think only upstream should be regenerated, in this case only B, or maybe I missing something?

> If A is not iterative, but B is iterative, we should regenerate A and B, but what about C and D?

How can B be iterative when all its dependencies are not? Maybe other dependencies that are not proper OpenPNM dictionary keys? Like scalars that, for some reason, might get updated during iterations?

I think, ideally, we only want to declare iterativeness of the bottom-most properties. For example, instead of setting `diffusive_conductance` to iterative, we'd rather set, for example, `viscosity` to iterative, and the algorithm automatically "infers" the iterativeness of upstream properties. Does that make sense?

jgostick:
My diagram was meant to show that B and C-D were both upstream.

Regarding B being iterative but A is not, I think this is what you're saying about conductance.  If viscosity is set to be iterative (because of changing temperature), then conductance should inherit this mode.  Which is what I mean by B is set to iterative (viscosity) but A is not (conductance).

ma-sadeghi:
This is addressed in PR #1445.

PR #1445, opened 2020-03-16
Algorithm dependency handling

ma-sadeghi:
This PR fixes issue #1393. We no longer need to manually store iterative properties in an algorithm's settings dictionary, the algorithm rather finds them on its own using the `dependency_graph` of the constituent objects (phase and physics for now, but could extend to network and geometry as well).


openpnm/core/ModelsMixin.py
```diff
@@ -45,10 +45,17 @@ def dependency_list(self):
         d = nx.algorithms.dag.lexicographical_topological_sort(dtree, sorted)
         return list(d)
 
-    def dependency_graph(self):
+    def dependency_graph(self, mode="shallow"):
```
jgostick: To be consistent with other methods, I think this arg should be ``deep=False``.  Since it's only a binary choice, accepting an open-end mode is confusing.  By using a flag, it is more obvious what it does.

ma-sadeghi: You're right, makes sense. It's fixed now.

jgostick:
This is amazing, but is kind of a big deal.  I think I'll wait until our Friday morning telecon to merge it, so you can walk me though it.

jgostick:
We need to re-add the ability for users to set iterative props, like 'pore.diameter', and than have this new method (find_iterative_props) resolve all it's dependencies as well.  So it should handle quantity as well as the items in the iterative_prop list.

PR #1472, opened 2020-04-21
Add permeability benchmark examples

ma-sadeghi:
This PR adds three (almost) identical examples, each for a different type of sandstone: (1) Berea, (2) Benthemeir, and (3) Doddington. The examples are in IPython notebook format, which are tested via `nbval`. So, next time we make some changes to `OpenPNM` that could potentially break our hydraulic conductance models in anyway (via shape factor models, etc.), hopefully we will know in advance.


ma-sadeghi:
Currently, I added both `.pnm` and `.dict` files for each sandstone since the `PoreSpy` IO class only accepts `.dict` at the moment. When we add support for `.pnm`, we can get rid of the redundant files.

PR #1488, opened 2020-05-25
PyPardiso wrapper + refactoring solvers

ma-sadeghi:
This PR adds a wrapper to PyPardiso solver. Also, it refactors our solver interface (although technically we don't have a proper solver interface yet). The refactoring is in the direction of creating a proper solver interface for OpenPNM in the future (ex. `Solver` class).


jgostick:
@ma-sadeghi You're going to have to walk me through this in a telecon.  I'm curious about this 'future' Solver class direction.

PR #1489, opened 2020-05-26
Fix bug in set_rate_BC

### LINKED ISSUE #1479: Problem with set_rate_BC when it gets array of pores and arrays of rate values
Using an array of values to set a rate BC for an array of pores (the same length) in ionic conductantion algorithm. For some reason it gives an error that goes back to generic transport _set_BC self['pore.bc_' + bctype][intersect] += values line. With error "operands could not be broadcast together with shapes (0,) (1808,) (0,)" .

ma-sadeghi:
This PR closes #1479.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,7 @@ codecov:
   branch: master
 
 coverage:
-  precision: 2
+  precision: 1
   round: down
   range: "50...100"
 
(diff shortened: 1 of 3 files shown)


Ni2M:
Thank you very much @ma-sadeghi . It's now working perfectly.

PR #1496, opened 2020-05-31
add axis' labels for `plot_coordinates` method

xu-kai-xu:
When I use quick_plot method to visualize the network, I can not distinguish the three axis. So I add `ax.set_xlabel('x-axis')` to show the axes for the quick plot method.


ma-sadeghi:
!runci

jgostick (changes requested):
This is a reasonable thing to add, thank you very much for this contribution.  Just a few comments:

Won't this break if the network is 3D?  These lines should be inside the 'else' statement on line 2271?  And for that matter, could you add the same for 3D graphs, but with a z-axis as well, inside the ``if ThreeD`` clause?

xu-kai-xu:
> This is a reasonable thing to add, thank you very much for this contribution. Just a few comments:
> 
> Won't this break if the network is 3D? These lines should be inside the 'else' statement on line 2271? And for that matter, could you add the same for 3D graphs, but with a z-axis as well, inside the `if ThreeD` clause?

I tried. But I can not find a way to add labels for all the three axes. If the network is 2D, only two axes, it is ok. If the network is 3D, only x and y axes are added labels. Then the one without labels is the z-axis. For example:

image

I will mark this problem and try to figure it out as soon as possible.

xu-kai-xu:
> This is a reasonable thing to add, thank you very much for this contribution. Just a few comments:
> 
> Won't this break if the network is 3D? These lines should be inside the 'else' statement on line 2271? And for that matter, could you add the same for 3D graphs, but with a z-axis as well, inside the `if ThreeD` clause?

Ok, I have made the changes and then it will show nice labels for both 2D and 3D networks.

ma-sadeghi:
Hey @xu-kai-xu. Sorry, I just saw your comment. I already made some changes. Please take a look. Thanks!

ma-sadeghi:
Here's the output of `plot_connections` and `plot_coordinates` for different network shapes (`Cubic`):
# 3D
- (a) [3, 4, 5]
image

# 2D
- (b) [3, 4, 1]
image

- (c) [3, 1, 5]
image

- (d) [1, 4, 5]
image

# 1D
- (e) [3, 1, 1]
image

- (f) [1, 4, 1]
image

- (g) [1, 1, 5]
image

xu-kai-xu:
> Here's the output of `plot_connections` and `plot_coordinates` for different network shapes (`Cubic`):
> 
> # 3D
> * (a) [3, 4, 5]
>   image
> 
> # 2D
> * (b) [3, 4, 1]
>   image
> * (c) [3, 1, 5]
>   image
> * (d) [1, 4, 5]
>   image
> 
> # 1D
> * (e) [3, 1, 1]
>   image
> * (f) [1, 4, 1]
>   image
> * (g) [1, 1, 5]
>   image

Nice!

jgostick (approved):
This looks great now, thanks @ma-sadeghi and @xu-kai-xu !

jgostick:
It's late so I'm going to wait until tomorrow to merge this, having learned my lesson last time!

PR #1509, opened 2020-06-07
Fixed bug in bond_percolation that allowed early invasion in some cases

jgostick:
This is a strange one.  

When networks have some throats that are larger than the pores they connect, the OP algorithm was marking them as invaded at pressure below their capillary entry pressure.  

I combed through the code and found a bug, which I fixed, and the issue with OP went away...the problem is that I don't know why this particular bug was being triggered!  And I don't know why it was not triggered when throats were all smaller than pores.

Anyway, I'll keep looking at it, and maybe  it will occur to me, while I write a test.


jgostick:
I think I get the issue now.  Consider the following network:

| 1 |--a--| 2 |--b--| 3 |--c--| 4 |

If throats a and c are occupied, then pores 1, 2 3 & 4 will be occupied.  The bond_percolation algorithm was inspecting the cluster number of pores 2 and 3, and assigning the minimum to throat b, so if throat b was not occupied it would mistakenly be given a cluster number, making it occupied and thus declaring these 4 pores as one big cluster.  This bug thus makes throat b appear occupied at a pressure below it's entry pressure. 

I still don't understand why it only appears when throats were bigger than pores.  Perhaps this is just an illusion and it was just statistically far less likely.

jgostick:
OK, I have finally figured out to my satisfaction what was going on here:

My troubles started because I could not reproduce this effect on normal stick and ball.   I thought it was because the throats were guaranteed to be smaller, but it's actually because the throats as are factor of the smallest neighbor. It's subtle, but this arrangement guarantees that a bond was never accidentally mislabeled as invaded too early.  It was just mathematically impossible due to the way throat sizes around a given pore were assigned. As soon as I added even a tiny amount of randomness to the sizes, the problem occurred.  Anyway, I am now happy that the 1-line fix is correct.  

In summary, this was a bad bug, but was somewhat hidden since it did not affect the pore invasion, only the throats, and the invasion patterns were not really impacted.  Basically, at each pressure step some throats were being added, extending out from the invading cluster, but not more than one step.  The access limitation check removed any clusters that would have been erroneously added via one of the bad throats.  We are very fortunate that this access limitation check worked the way it did or else massive errors would have crept in.  I should conceded that any simulations done without access limitations would be very badly off, but I don't think this is a very common use case.

1minus1:
Thanks for chasing this down, @jgostick!

1minus1:
Just took it for a spin, and it looks good.

Before:
image

After:
image

The network invasion looks much more correct after the fix, as well.

image

PR #1511, opened 2020-06-09
Bugfix: variable conductance now triggers A rebuild

ma-sadeghi:
Major changes:
- `solver_maxiter` renamed to `solver_max_iter`
- `max_iter` renamed to `nlin_max_iter` for clarity
- `_find_iterative_props` renamed to `_get_iterative_props` for consistency
- `A` is forced to be rebuilt for variable `settings["conductance"]`
- Effective diffusivity calculation was added to benchmark examples
- `GenericTransportSettings` and `ReactiveTransportSettings` docstrings massively beefed up

Here's a sample script that describes simple Fickian diffusion but with variable diffusivity, such that diffusivity is significantly higher at high concentrations:

```python
import numpy as np
import openpnm as op
import matplotlib.pyplot as plt

np.random.seed(10)
ws = op.Workspace()


def variable_diffusivity(target, c_ref, pore_concentration="pore.concentration"):
    X = target[pore_concentration]
    val = 1e-9 * (1 + 5 * (X / c_ref) ** 2)
    return val


net = op.network.Cubic([1, 15, 15], spacing=1e-5)
geom = op.geometry.StickAndBall(network=net)
air = op.phases.Water(network=net)
air.add_model(propname="pore.diffusivity", model=variable_diffusivity, c_ref=10.0,
              regen_mode="deferred")
phys = op.physics.Standard(network=net, geometry=geom, phase=air)
fd = op.algorithms.FickianDiffusion(network=net, phase=air)
fd.set_value_BC(pores=net.pores("left"), values=10.0)
fd.set_value_BC(pores=net.pores("right"), values=0.0)
fd.run()

c = fd["pore.concentration"]

plt.style.use(["science", "ieee", "amin"])
fig, ax = plt.subplots()
c_xz_avg = c.reshape(net.shape).mean(axis=(0, 2))
ax.plot(c_xz_avg, "ko-")
ax.set_xlabel(r"x")
ax.set_ylabel(r"c")
plt.show()

ws.clear()
```
Here's the plane-averaged concentration profile before the fix:
image

and here's after:
image


jgostick (changes requested):
* I think the ``set_variable_properties`` method should be made private for now.  It's not really needed, so let's hide it, and either we think of a use case or delete it in future.
* If we're going to change the settings names, I'd suggest we maybe have a look at some of the others too.  This is a breaking change, so might as well really break it.

PR #1513, opened 2020-06-11
Examples galore

### LINKED ISSUE #1467: Update readme to suggest 'watching' this repo to be notified of updates
It is a mystery to me why github makes is so hard to mass communicate with users about updates.  Starring the repo is not enough to get notified of new releases.  It is necessary to actually 'watch' it.  Then you can go into your github settings and enable email notifications on new releases (this is actually the default behavior).  Anyway, I think we should start steering people toward starring AND watching the repo.

### LINKED ISSUE #1499: Revise docs for website
The docs require that pages be added for each algorithm, and that these pages be linked to in the index.rst file.  We have fallen behind on this for the new PNP algorithms.  

We should for sure update this asap, BUT, maybe we can find a way to do this programatically to ensure that every new class/feature we add is automatically included?

### LINKED ISSUE #1515: Update examples
I have started a new branch called 'examples_galore' and I plan to add and update many examples in the near future.  I am opening this issue as a single place to put all the little issues that I find, instead of opening dozens of new issues in the bug tracker.  I'll use @ mentions to get attention of people as needed.

- [ ] Using PNP algorithm (@magnaou)
- [ ] Using Salome IO class (@magnaou)
- [ ] Fix Blunt et al. example on Pore-Scale Imaging (?)
- [ ] Fix Wu et al. example (?)
- [ ] Showing the difference between `OrdinaryPercolation` and `InvasionPercolation`
- [ ] Generating hierarchical networks (@ma-sadeghi)
- [ ] Update examples to remove the `net.update(snow)` instruction (@Zohaib-Atiq)
- [ ] Labels, their meaning, creation, and use, including the new ``set_label`` method (I thought we had this, but I couldn't find it)
- [ ] Adjusting pore size distributions to match MIP data
- [ ] Use of VoronoiDelaunay class (it's well described in a visualization example, but does not have a stand alone example)

jgostick:
I plan to add several examples on this PR, since we seem to be lagging behind in this area lately.

Fixes #1467
Closes #1515
Fixes #1499


jgostick:
Also fixes #1467

PR #1514, opened 2020-06-12
add labels for 2D and 3D quick ploting.

xu-kai-xu:
@jgostick, as you suggested, add labels for both 2D and 3D network quick plotting method.
For example:

image


ma-sadeghi:
This is duplicate of #1496, so I'm closing it.

PR #1539, opened 2020-06-24
Readded color to plot_connections

jgostick:
It seems that the switch to LineCollection lost all the formatting that was passed in kwargs.  I have re-worked the function to work with the new formatting arguments required by LineCollection, and made some allowances to still work with the old args (only 'c', 'color' and 'alpha').


jgostick:
Strange test failure, seems to be deep inside scipy.sparse...must be us sending a wrong argument?

jgostick:
I have confirmed that this is a scipy problem.  Amin ran tests yesterday using scipy 1.4.1 and they passed.  My tests ran later in the day, using scipy 1.5.0, and they failed.  I just re-ran Amins tests, which today used 1.5.0, and they now fail. 

I've dug into the scipy.sparse source code, and they've changed something major.  In fact, the function that fails in 1.5 does not even exist in the 1.4 code.  (lil_flatten_to_array)

I will open an issue over there.

jgostick:
apparently it was not a bug per-se, but we were actually doing something with LIL arrays that was not allowed.  I have updated the code in the iMorph file (the ONLY place we were breaking it), so this will be fixed when we merge this PR.

ma-sadeghi (approved):
Looks good. I just made a little change to reuse `matplotlib`'s `to_rgb` method.

PR #1550, opened 2020-07-01
Removing conversion between cached formats in get_adjancency_matrix function

jgostick:


ma-sadeghi:
@jgostick Can this behaviour possibly be tested via unit tests?

PR #1559, opened 2020-07-03
Fixing stitch

### LINKED ISSUE #1532: stitch needs better handling of labels
I would be nice if you could specify the labels of the new stitched throats.

### LINKED ISSUE #1533: bug in stich when called repeatedly?
When I played with this it seemed like calling stitch consecutively deleted the 'throat.stitch' label from previous calls, another user did not see this problem.  Worth investigating.

jgostick:
Fixes #1532 and closes #1533 which does not seem to be a problem.

the stitch function now: 
* accepts ``label_stitches`` which will apply a list of specified labels to any newly generated throats
* has a ``radius`` method which only connects to pores within the given ``len_max`` distance
* has an improved ``nearest`` method which actually does what it says and only connects to the true nearest pore

This is actually a breaking change since the 'radius' method now acts like the 'nearest' method used to, and the 'nearest' method ignores the ``len_max`` argument

### DIFF
--- a/openpnm/network/Bravais.py
+++ b/openpnm/network/Bravais.py
@@ -155,13 +155,13 @@ class Bravais(GenericNetwork):
                 n.update({'throat.all': np.array([], dtype=bool)})
                 n.update({'throat.conns': np.ndarray([0, 2], dtype=bool)})
             # Join networks 2, 3 and 4 into one with all face sites
-            topotools.stitch(net2, net3, net2.Ps, net3.Ps,
-                             len_min=0.70, len_max=0.75)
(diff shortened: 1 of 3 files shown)


jgostick:
@ma-sadeghi I broke some backwards compatibility on this one, as mentioned above, so I thought I'd let someone else look at it and see if it's ok.

ma-sadeghi (approved):
Looks good, and I'd say this is a minor backwards-incompatible change, so maybe we shouldn't worry about it.

ma-sadeghi:
@jgostick Btw, you can use "Enhanced" keyword in merge messages too (for release notes generator).

PR #1563, opened 2020-07-07
Speed up find_neighbor_sites and num_neighbors

ma-sadeghi:
**Optimized**
```python
Flatten: True
num_neighbors:       720 ms ± 7.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
find_neighbor_sites: 716 ms ± 8.64 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Flatten: False
num_neighbors:       219 ms ± 1.14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
find_neighbor_sites: 2.73 s ± 12.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
**Dev**
```python
Flatten: True
num_neighbors:       2.87 s ± 14.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
find_neighbor_sites: 2.97 s ± 84.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Flatten: False
num_neighbors:       5.88 s ± 67.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
find_neighbor_sites: 4.9 s ± 64 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
- With `flatten=True`, they're both **~4x** faster.
- With `flatten=False`, `num_neighbors` is **~30x** faster, and `find_neighbor_sites` is **~2x** faster.

### DIFF
--- a/examples/tutorials/Intro to OpenPNM - Intermediate.ipynb
+++ b/examples/tutorials/Intro to OpenPNM - Intermediate.ipynb
@@ -55,14 +55,14 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "pn = op.network.Cubic(shape=[20, 20, 10], spacing=0.0001, connectivity=8)"
+    "pn = op.network.Cubic(shape=[20, 20, 10], spacing=0.0001, connectivity=14)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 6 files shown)


jgostick (approved):
This looks great...I guess my use of numba on the ``speedup_num_neighbors`` branch was the wrong approach?

ma-sadeghi:
@jgostick I ended up not using `numba` all. I tried a bunch of different approaches and none worked.

PR #1566, opened 2020-07-07
Enhance ball_and_stick flow shape factor model when L/D > 0.5

### LINKED ISSUE #1421: flow_shape_factor models yield nans when L / D > 0.5
[The problem is on this line](https://github.com/PMEAL/OpenPNM/blob/b3873d35270b0acaad019264368d0055c677d159/openpnm/models/physics/flow_shape_factors.py#L99)

The arctan term goes to inf when the argument == 1 and nan when arg > 1, so when L / D > 0.5 it's then multiplied by 2, and creates an undefined result.  This is putting nans in the hydraulic conductance array.

Ni2M:
Integral of (1/A2) is calculated for a pore with verying cross section of circular shape. Fixes #1421


jgostick:
Since this error seems to be fundamental, there is no 'fix'.  Instead, we should do a check for L/D > 0.5 and throw an exception with a clear message to the user.

PR #1571, opened 2020-07-11
Hotfix: caching of A only worked for ReactiveTransport, but not other subclassed algorithms

### LINKED ISSUE #1548: Issue with alg._A not updating properly
I wrote a script where I run transient fickian diffusion algorithm. My problem required that I stop the solver to update models including diffusive_conductance model and then restart solver using last solution as new initial condition. While my diffusive_conductance values were updated properly when I printed alg._A I could see that the A matrix wasn't being updated properly. I worked around this issue by recreating the algorithm object on every iteration. @ma-sadeghi thought that I shouldn't have to recreate the algorithm object for this to work.

ma-sadeghi:
Fixes #1548. 

I thought we had already fixed this, and indeed we had done that in PR #1511, but the fix only worked for `ReactiveTransport`, but none of the subclassed methods which overloaded the `run` method, such as `NernstPlanck`, `TransientIonicTransport`, etc.

This PR addresses the issue at much more fundamental level, which is inside `GenericTransport._build_A` and therefore, it pretty much covers any algorithm that relies on `_build_A` (which would be most of our algorithms).

## What it does?
It resets `cache_A` and `cache_b` to `False` if conductance quantity (e.g. `throat.diffusive_conductance`, etc.) is found in `_get_iterative_props()`, which forces `_build_A` to re-create `A` every time it's called.


jgostick (approved):
Looks good, but should we merge this into release and push a patch to pypi?   A bit embarrassing to do it 1 day after a major release, but this seems worth pushing asap.

ma-sadeghi:
> Looks good, but should we merge this into release and push a patch to pypi? A bit embarrassing to do it 1 day after a major release, but this seems worth pushing asap.

I agree, it's worth pushing. Should we merge into dev and then dev into release?

jgostick:
either way...there is nothing new on dev, but I guess it's a bit easier so we don't have to cherrypick the merge back on to dev.

PR #1576, opened 2020-07-13
Significant change to datatype conversion in interleave_data

### LINKED ISSUE #1574: [Bug]: <Assigning integer value to geo2['pore.diameter'] changes all values in pn['pore.diameter'] to that integer>
**Describe the bug**
After a geometry is updated using op.geometry.Imported(net) and new boundary pores are added, when assigning attributes to the new boundary pores if the value is an integer it changes all values of that attribute to the integer. If a float is used no issue. Found in https://github.com/PMEAL/OpenPNM/blob/dev/examples/extractions/Managing%20Geometrical%20Properties%20of%20Imported%20Networks.ipynb (CELL 16) geo2['pore.diameter '] = 2.0 VS. geo2['pore.diameter '] = 2. (using 2 overwrites the network pore.diameter values) 

**To Reproduce**
Steps to reproduce the behavior, ideally a code snippet that only reproduces the bug:
```python
<
pn = op.network.Delaunay(shape=[1, 1, 0], num_points=100)
pn['pore.diameter'] = np.random.rand(pn.Np)
print(pn)

geo = op.geometry.Imported(network=pn)

print(geo)

op.topotools.extend(network=pn, pore_coords = [[1.2, 1.2, 0]], labels='new')

geo2 = op.geometry.GenericGeometry(network=pn, pores=pn.pores('new'))
geo2['pore.diameter'] = 0

print(geo2)

print(pn['pore.diameter'])>
```

**Expected behavior**
The final print command should print the pn['pore.diameter'] values but instead it prints a list of 0s the length of geo + geo2.

**Screenshots**
image
vs. 
image

**Additional context**

jgostick:
Fixed #1574

Data types were not being inferred properly when doing interleave data.  The new method does a check for simple cases first, then goes onto the complex logic if needed.


ma-sadeghi (approved):
Looks good!

PR #1582, opened 2020-07-16
Cone model for hydraulic conductance

magnaou:
Cone model for hydraulic conductance.


ma-sadeghi (changes requested):
@magnaou Can you please add unit tests for this?

magnaou:
yes, sure

ma-sadeghi (dismissed):
@magnaou Looks great.

jgostick:
How does this PR affect the conical frustrum that Niloo is working on, as per #1421

magnaou:
> How does this PR affect the conical frustrum that Niloo is working on, as per #1421

This model shouldn't take any shape factor as we discussed before.

ma-sadeghi:
@jgostick @magnaou Shouldn't this be a shape factor model, rather than a conductance model?

magnaou:
No, it is a formal derivation of an analytical solution of a Stokes flow problem in a cone. Equivalent to Hagen-Poiseuille but in a cone instead of a cylinder.

ma-sadeghi:
> No, it is a formal derivation of an analytical solution of a Stokes flow problem in a cone. Equivalent to Hagen-Poiseuille but in a cone instead of a cylinder.

Yes, but other shape factor models are the same. They're formal derivation of analytical solution of Stokes flow for a particular cross-section. For example, `ball_and_stick` flow shape factor model is assuming spherical shape for pores and cylinders for throats.

magnaou:
> > No, it is a formal derivation of an analytical solution of a Stokes flow problem in a cone. Equivalent to Hagen-Poiseuille but in a cone instead of a cylinder.
> 
> Yes, but other shape factor models are the same. They're formal derivation of analytical solution of Stokes flow for a particular cross-section. For example, `ball_and_stick` flow shape factor model is assuming spherical shape for pores and cylinders for throats.

I didn't thoroughly investigate what we have on the "shape_factor" models for flow. I know what they do, of course. However, the solutions I get using these models is different from what I get with the cone model in this PR. Which means they are different? I also have comparisons with experimental data and DNS on sphere packs and this cone model is giving the best accuracy. I can show you the results if you are interested. It's part of the collaboration with Antonio.

magnaou:
Don't merge yet. To be discussed with @jgostick and @ma-sadeghi during the marathon or before.

ma-sadeghi:
> > > No, it is a formal derivation of an analytical solution of a Stokes flow problem in a cone. Equivalent to Hagen-Poiseuille but in a cone instead of a cylinder.
> > 
> > 
> > Yes, but other shape factor models are the same. They're formal derivation of analytical solution of Stokes flow for a particular cross-section. For example, `ball_and_stick` flow shape factor model is assuming spherical shape for pores and cylinders for throats.
> 
> I didn't thoroughly investigate what we have on the "shape_factor" models for flow. I know what they do, of course. However, the solutions I get using these models is different from what I get with the cone model in this PR. Which means they are different? I also have comparisons with experimental data and DNS on sphere packs and this cone model is giving the best accuracy. I can show you the results if you are interested. It's part of the collaboration with Antonio.

I didn't mean that this is identical to other shape factors. I meant this should a shape factor model, and not a conductance model. If I remember correctly, the way our shape factors are set up is such that the baseline is assuming cylinders in series, and the shape factor needs to correct for that baseline.

magnaou:
> > > > No, it is a formal derivation of an analytical solution of a Stokes flow problem in a cone. Equivalent to Hagen-Poiseuille but in a cone instead of a cylinder.
> > > 
> > > 
> > > Yes, but other shape factor models are the same. They're formal derivation of analytical solution of Stokes flow for a particular cross-section. For example, `ball_and_stick` flow shape factor model is assuming spherical shape for pores and cylinders for throats.
> > 
> > 
> > I didn't thoroughly investigate what we have on the "shape_factor" models for flow. I know what they do, of course. However, the solutions I get using these models is different from what I get with the cone model in this PR. Which means they are different? I also have comparisons with experimental data and DNS on sphere packs and this cone model is giving the best accuracy. I can show you the results if you are interested. It's part of the collaboration with Antonio.
> 
> I didn't mean that this is identical to other shape factors. I meant this should a shape factor model, and not a conductance model. If I remember correctly, the way our shape factors are set up is such that the baseline is assuming cylinders in series, and the shape factor needs to correct for that baseline.

Ok, I see what you mean @ma-sadeghi . In this PR, the baseline is a cone (not a cylinder) and shape factors are 1 (i.e., no shape factor correction). To be discussed during a teleconference I guess...

ma-sadeghi:
Being addressed in PR #1604

PR #1590, opened 2020-07-21
Effective volume

### LINKED ISSUE #1459: Throat volume is not accounted for in TransientReactiveTransport class
In `TransientReactiveTransport` class, under `_t_update_A` and `_t_update_b` methods, we only account for `pore.volume` and not `throat.volume`. We should distribute `throat.volume` based on `pore.volume`. Should be an easy fix, but important one, especially when `throat.volume` is large enough.

mkaguer:
Fixes #1459


ma-sadeghi:
@mkaguer You can mention the issue this PR is closing/fixing by writing "fixes #number_of_issue" (I already did it for you). This way, the issue automatically gets closed when we merge this PR and there's no need to mention the PR in the issue page either.

mkaguer:
Okay, I see. Thanks @ma-sadeghi !

ma-sadeghi (changes requested):
Great job Mike! Only two things:
- Some tests are failing. We should make sure all tests are passing so our changes won't break the existing code. In this case, it's not breaking any code, it's just PEP8 warnings. PEP8 is a agreed-upon style guide for Python code, ex. variable names should be lower cased, etc. To see which tests are failing, go to the "Conversation" tab of this PR, locate which tests are failing, click on the failing tests and scroll till the end where it says: "FAILURES", and try to fix the error(s). In your case, the errors are just missing white space, so nothing major, but still needs to be fixed before merging.
- Whenever we add "new" code, we should ideally add a "unit test" to test its behaviour to make sure it's doing what it's supposed to. For instance, in your case, you need to create a very small network, I'd say 2 by 2, and test your "effective" model and compare the results with hand calculations. To see how to write unit tests look under `OpenPNM\tests\unit\models\geometry` and open `PoreVolumeTest.py` and add your own test similar to other tests you find in that file. I'm sure you can figure it out yourself, but let me know if you need a hand.

ma-sadeghi (approved):
Look great! Nice job @mkaguer!

ma-sadeghi (changes requested):
@mkaguer One last thing: you need to filter out throats that are relevant to the `target` object. If `target` spans the entire network, your model works fine, but if `target` is part of your entire network, it fails. To elaborate, this line:
```
np.add.at(eff_vol, P1, 1/2*target[throat_volume])
```
adds some stuff to `eff_vol` in locations `P1`. However, if your network has two geometries, then `target` doesn't span the entire network and therefore, some of `P1` locations are not in `eff_vol`. See one of `throat_length.py` models to get the idea.

ma-sadeghi (approved):

PR #1604, opened 2020-08-12
New conduit flow approach

### LINKED ISSUE #1308: Conductance models should optionally return element and conduit values
It might be a good idea to add a ``mode`` argument to conductance models, that accepts 'elements' and 'conduit'. 

In the case of 'conduit' the model returns a single conductance value for the whole conduit.  In the case of 'elements' it returns a Nt-by-3 array or dict containing the conductance of each individual element in the conduit.  This latter option would be consistent with our 'conduit.length' and 'conduit.area' geometry models.

jgostick:
Closes #1308. I have created a 'proof-of-concept' for this new shape factor idea.  There is a script added to the top level that illustrates it. 

# To Think About
* Is it possible to adopt the new approach, without breaking things?  Like is there a way to use Standard physics with the new size factors?
* How can 'aspect ratios' be used in a general way? Can we get them from network extraction and also be consistent with a generated network of 'flattened' cubes?  

# To Do
- [x] Roll back changes to StickAndBall.  This should still work the 'old' way (for now at least)
- [x] Create several new Geometry classes
  - [x] `SpheresAndCylinders`
  - [x] `ConesAndCylinders`
  - [x] `PyramidsAndCuboids`
  - [x] `CubesAndCuboids`
  - [x] `CirclesAndRectangles`
  - [x] `TrapezoidsAndRectangles`
  - [x] `SquaresAndRectangles`
- [x] Update the Imported geometry class to use these new conductance models, utilizing the settings to control pore and throat shapes
- [x] ~Fix-up `throat_endpoints` models~ Add `conduit_lengths.py` module (shortcut for `throat_endpoints` -> `conduit_lengths`)
  - [x] ~Refactor and rename existing models~
  - [x] Add new models to service the size_factor models
- [x] Finalize `hydraulic_size_factors` file
  - [x] `spheres_and_cylinders`
  - [x] `circles_and_rectangles`
  - [x] `cones_and_cylinders`
  - [x] `trapezoids_and_rectangles`
  - [x] `cubes_and_cuboids`
  - [x] `squares_and_rectangles`
  - [x] `ncylinders_in_series`
  - [x] ~Pull code from Mehrez's cone model in PR #1582~ (this should be moved to a new PR)
- [x] Finalize `diffusive_size_factors` file
  - [x] `spheres_and_cylinders`
  - [x] `circles_and_rectangles`
  - [x] `cones_and_cylinders`
  - [x] `trapezoids_and_rectangles`
  - [x] `cubes_and_cuboids`
  - [x] `squares_and_rectangles`
  - [x] `ncylinders_in_series`
- [x] Roll back changes to `Standard` physics
- [x] Add unit tests for `conduit_lengths` and size factor models


ma-sadeghi:
@jgostick Nice.

Now that I'm rethinking it, here are some things to consider (although I guess they were all brought up in a sense in our yesterday meeting):
- Instantiating a geometry model from within a physics model doesn't seem right.
- Conductance models (`stokes_generic`, etc.) are doing very little now, basically multiplying the flow coefficient by a scalar (viscosity). Are they really worthy?
- How about not moving conductance models to geometry, keeping them where they were, but specializing them via wrappers. So, pretty much what you've done here, except move the "flow coefficient" calculations inside the specialized models, such as `stokes_conical_frustum`.

I know that "flow coefficient" is entirely geometric, but I'm wondering if the added complexity due to this decoupling is really worth it.

We had a similar situation with Knudsen diffusivity. Knudsen diffusion has two components: an entirely geometric part and another relating to bulk diffusion coefficient. We went back and forth a bunch of times but in the end decided to keep it as a physics model.

Anyways, just some thoughts. What do you think? @jgostick

jgostick:
@ma-sadeghi I think I agree with you on most counts, but there are a few 'upsides' to this new approach:

1. Have these 'size_coeff' models on geometry is actually great for 'materials', like voronoi and Berea, which return a project containing a network and a geometry.  It's very handy to apply the models at this stage, as a convenience to the user, instead of expecting them to figure out which models to apply when they create their own physics.  
2.  The same logic applies to imported networks, which will have the AI generated 'size_coeff' as a geometry property. In fact this is pretty much unavoidable, so regardless of how we handle the models like 'conical_frustrum', we still need that 'generic_stokes' model to do the trivial calculation of 'size_coeff / viscosity'.  So given that we actually need to to add this approach for imported networks, maybe we should support it for other situations as well.

I'm undecided.  I'm working on a flow diagram that can hopefully help us visualize the situation.

Ni2M:
As the diffusive_shape_factors in the physics model will be deleted later, we include the calculation of the shape factors in the geometry diffusive_coefficient model https://github.com/PMEAL/OpenPNM/blob/e666e583cb3c1353f52bb844365ce4556635af0a/openpnm/models/geometry/conduit_diffusive_coefficient.py#L60   instead of calling [diffusive_shape_factors method](https://github.com/PMEAL/OpenPNM/blob/new_conduit_flow_approach/openpnm/models/physics/diffusive_shape_factors.py).  Because diffusive_shape_factors method works with the phase/physics whereas in geometry model we need to find the g_vals before assigning any physics/phase to the problem.

Ni2M:
```python
import openpnm as op
net=op.network.Cubic(shape=[10,10,10])
geo=op.geometry.StickAndBall(network=net)
diff_SC=geo['throat.diffusive_shape_coefficient']
# array([0.07231972, 0.06807391, 0.04211285, ..., 0.03845162, 0.05079792, 0.05877204])
```
works properly for this section. It gives the diffusive_shape_coefficient (1 value as return_elements=False by default)
by default we set the StickAndBall geometry's diffusive_shape_coefficient model as spheres and cylinders.
For an approximation of the throat and pore length the calculation from the a,r (if the conduit length is to given to the model) https://github.com/PMEAL/OpenPNM/blob/3a7f7075fd4bcf3f781c44b06218e6b7e3b8e8cb/openpnm/models/geometry/conduit_diffusive_coefficient.py#L47 is problematic. I will look into it in more details.

Ni2M:
[Dev meeting Oct8.pptx](https://github.com/PMEAL/OpenPNM/files/5349574/Dev.meeting.Oct8.pptx)

Ni2M:
[Dev meeting Oct9.pptx](https://github.com/PMEAL/OpenPNM/files/5349578/Dev.meeting.Oct9.pptx)

Ni2M:
After some thoughts and trying the codes on two structures of the conduit_coefficient and stick and ball geometry, there were some problems in structure1 that I continued based on the structure2 instead:

structure1: remove all properties that require the shape (pore.volume, throat.volume, conduit length,...) form the stick and ball geometry class because those features are only related to sphere_and_cylinder assumption not cone, etc. This makes problem when the user wants to do some other calculation that require pore.volume, throat.volume, etc. Then they have to define those even after calling stick and ball geometry. another problem is if we remove them from the geometry and add them to the sphere_and_cylinder conduit_coefficient method to internalize those calculation, it requires adding models to the network (add geometrical models for the endpoints, throat length, pore volume,..). Adding those model within a model is problematic, spcecialy in this case where we are adding a geometry model to the network (not geometry object). If we instead consider adding the codes for calculation of the endpoint and throat length, etc in the conduit_coefficient methods, that will add extra amount of duplicated codes.
strucuture2: If we keep the pore.volume, conduit length, .. models in the stick and ball geometry, it will save us from the problems in structure1. The downside is the lengths in each diffusive_coefficient method is different. Those length that we found from the conduit length at the beginning in geometry might need to be overwritten after we call the conduit_coefficient methods.

Ni2M:
In hydraulic conductance file, these methods are not consistent with the new approach ( if we want to consider more models than only hagen_poiseuille). These methodes can not be called from the stokes generic method, as they have a different structure of finding g and shape factors. We should find a solution to include them or maybe change them in a consistent way.
- hagen_poiseuille_power_law
- valvatne_blunt

A similar problem exists for the diffusion conductance. Any method other than the ordinary diffusion can not follow the new structure. 
-  mixed_diffusion
- taylor-aris
- multiphase

Ni2M:
Planning: 
- move the shape factor models to geometry instead of physics
- call those models within their corresponding conduit_diffusive_coefficient method 9sphere and cylinder,..) to get rid of the extra lines and further maintenance of the code
- remove the wrappers and methods in diffusive and hydraulic conductance, keep the generic ones and the rest of the methods

ma-sadeghi:
@Ni2M It might also be a good idea to hide them with a leading underscore to keep the namespace clean.

Ni2M:
@ma-sadeghi Right. I agree. I will follow that for the changes.

Issue #1570, opened 2020-07-11
Make conical_frustum_and_stick the default shape factor model


ma-sadeghi:
The default flow/Poisson shape factor models in `Standard` physics assume spherical pores. This assumption is not a robust one in terms of numerical calculations (see #1421 for instance). If we use `conical_frustum_and_stick` shape factor models instead, we won't have any issues with the values becoming `inf` or `nan` as they're mathematically relatively simple. Also, while the spherical assumption is best for an actual stick and ball geometry, it's not realistic in other cases and in my experience (if I remember correctly), `conical_frustum_and_stick` shape factor models overall give more accurate results.

jgostick:
We have actually side-stepped this one by creating the new size-factor models in PR #1604

PR #1605, opened 2020-08-12
Enhance ball_and_stick flow shape factor to raise Exception when pore diameter is invalid

### LINKED ISSUE #1421: flow_shape_factor models yield nans when L / D > 0.5

Ni2M:
Fixes #1421. Raise Exception for pores that can't be modeled with stick and ball shap factor model.
For pores with Dp <= 2*Lp the stick and ball shape factor model is not valid (This interferes with the assumptions and physical representation of the stick and ball). Use other models instead.


ma-sadeghi (changes requested):
Good job. Looks good @Ni2M. I have a few comments:

1. Could you please also add a unit test to capture the exception?
2. Also, I think we should do the same for poisson shape factor as well.
3. Use `np.sum` instead of Python's `sum`, it's much faster ([ref](https://stackoverflow.com/questions/10922231/pythons-sum-vs-numpys-numpy-sum)):
image

Ni2M:
Thank you @ma-sadeghi for your guidance. Yes, Sure.

Ni2M:
The changes lead to error in Standard physics definition https://github.com/PMEAL/OpenPNM/blob/3d1d1b0fd7bf4de610eda8c6377be1e113d02c10/openpnm/physics/Standard.py#L50 , which has ball_and_stick flow SF as default.  The error in [example paper recreation](https://github.com/PMEAL/OpenPNM/blob/dev/examples/paper_recreations/Gostick%20et%20al.%20(2007)/Gostick%20et%20al.%20(2007).ipynb) is related to this part as well, because it's using standard physics models.

ma-sadeghi:
> The changes lead to error in Standard physics definition
> 
> https://github.com/PMEAL/OpenPNM/blob/3d1d1b0fd7bf4de610eda8c6377be1e113d02c10/openpnm/physics/Standard.py#L50
> 
> , which has ball_and_stick flow SF as default. The error in [example paper recreation](https://github.com/PMEAL/OpenPNM/blob/dev/examples/paper_recreations/Gostick%20et%20al.%20(2007)/Gostick%20et%20al.%20(2007).ipynb) is related to this part as well, because it's using standard physics models.

As proposed by #1570, please go ahead and change the default shape factors (both flow and poisson) in `Standard` physics to `conical_frustum_and_stick`.

Ni2M:
Thank you @ma-sadeghi for your help. For sure.

ma-sadeghi (changes requested):
Great work @Ni2M. All tests are passing except for some examples. I guess it's because we've changed the default shape factor models in Standard physics. Here's how you can fix it:

Go to the Actions page and see which examples failed to run. Open them up locally in Jupyter Notebook, re-run all the cells, and "save" the file and commit the changes.

Ni2M:
Thank you @ma-sadeghi for your help. That is right the examples are making problem because of the changes in the values gives different results for permeability. I will follow your instructions to fix them.

Ni2M:
There is still difference between the number of pores and subsequent results in Simulation Berea example. I checked to make sure the PoreSpy is installed on its dev branch. New local run of the network doesn't match with the git run of the network.

ma-sadeghi:
> There is still difference between the number of pores and subsequent results in Simulation Berea example. I checked to make sure the PoreSpy is installed on its dev branch. New local run of the network doesn't match with the git run of the network.

Actually, you should use PyPI version of porespy, not the dev branch, as CI installs porespy using PyPI

Ni2M:
I used PyPI version of PoreSpy. The results of the extracted network in Berea_effective_prop example has 1 additional pore in contrast to the results of the CI run.

ma-sadeghi (approved):

ma-sadeghi:
Only one of the examples is failing, but that was discussed in our meeting, so merging it.

PR #1612, opened 2020-08-14
Fix plot_connections and plot_networkx scale

### LINKED ISSUE #1592: [Bug]: plot_connections and plot_networkx produce out-of-scale plots
**Describe the bug**
plot_connections and plot_networkx produce out-of-scale plots

**To Reproduce**
Steps to reproduce the behavior, ideally a code snippet that only reproduces the bug:
```python
import openpnm as op
net = op.network.Cubic(shape=[3, 3, 1])
op.topotools.plot_networkx(net)
op.topotools.plot_connections(net)
```

**Expected behavior**
Similar plot but with better margins.

**Screenshots**
Here's what you get:
image

Ni2M:
This PR fixes #1592 the problem with scale of network plots, where in some cases the network did not completely fit in the plot scale.


Ni2M:
Added feature: Previously, the plot_net only worked with spacing=1 for networks without geometry (to scale the node size with diameter). The effect of spacing#1 for networks without geometry is added to be compatible with the scaled node sizes. Y Dimensions are calculated as : Y[i]=(i+1)*spacing-0.5*spacing. So normalizing them by spacing fixes the overlap problem in plotting the connections. Each dimension is divided by its spacing (in case Xspacing#Yspacing).

Ni2M:
I think this PR is ready to merge.

ma-sadeghi (commented):
@Ni2M Can you upload the figures produced by each method (networkx and connections) for a 10 by 5 and a 10 by 10 2D cubic network?

Ni2M:
@ma-sadeghi  Thank you, for sure.

Ni2M:
**10 by 5 2D Cubic net**
plot_networkx
10_by_5

plot_connections
10_by_5_net

**10 by 10 2D Cubic net**
plot_networkx


plot_connections
1010_im

ma-sadeghi (changes requested):
`plot_connections` seems be working fine, except for the 10 by 10 case where the margins are still not quite right. But, the ones generated by `plot_networkx` don't look quite right, the spacing along x and y look noticeably different. 

I think you should revert the aspect parameter back to "equal", rather than "auto". That seems to be causing it.

Btw, did you try `plt.autoscale`?

Ni2M:
That is right. Sorry. I initially thought the spacing that it shows was not important to consider and I used "auto" mode so that the image is zoomed out to the x and y extent. I'll try it again with different modes. The only time I considered spacing was for improving its method to be able to plot connections and network for a network with spacing other that 1.

I used fig.tight_layout instead with a padding that was recommended in references. I changed it to autoscale now that you mentioned. It works better. thank you.

Ni2M:
**10 by 5 2D Cubic net**
network
105_n
connections
105_p
**10 by 10 2D Cubic net**
network
1010_n
connections
1010_p

Ni2M:
Note: Initially the code was developed for plotting networks with only default spacing (1). With some changes it is now capable of plotting neworks with different spacing. Note that in these cases plot_networkx still shows equal spacings, but plot_connections is consistent with different spacings.

ma-sadeghi:
> Note: Initially the code was developed for plotting networks with only default spacing (1). With some changes it is now capable of plotting neworks with different spacing. Note that in these cases plot_networkx still shows equal spacings, but plot_connections is consistent with different spacings.

Looks good, I just made a little change since not all networks have `spacing` as attribute (Voronoi, etc.).

ma-sadeghi (approved):

Ni2M:
Thank you @ma-sadeghi  for the changes. You are right about the spacing. That looks good now. Thank you.

PR #1621, opened 2020-08-20
New file format

### LINKED ISSUE #1619: All classes should be instantiable with no args
While working on the new file format, I'm wishing / needing that all classes can be created without passing in any args.  For instance, I want to create an empty Cubic class, but this is impossible.  I could create non-empty networks then delete the data afterwards, but this would require writing a custom delete sequence for every class, including each new one!  Ouch.

### LINKED ISSUE #1637: All IO classes should have a save_project method
At the moment there are some inconsistencies in the various apis of the IO classes.  I don't think we can start to rip these apart yet, but we can add a ``save_project`` method to all classes, and this should ONLY accept a project object, not a network or list of phases.  It may also be useful to allow any list of objects to be passed as a ``project``, so that users can save a specified set of objects if they wish.  I think this can be done in a backwards compatible way.

### LINKED ISSUE #1638: The IO classes should differentiate between 'save_project' and 'export_data'
Use of these terms would make it crystal clear what's going to happen.  We also need to banish the 'save' and 'load' terms, since they're too vague.  This would be backwards incompatible though.

jgostick:
Attempting to define a new file format for saving `OpenPNM` projects without using pickle so they are compatible across different versions. 

List of changes:
1. `pores` and `throats` are no longer optional for geometry classes
2. All network classes need to be instantiable without any argument at all, and there is a test for this
3. `save` and `load` methods are deprecated in favor of `export_data`/`import_data`
4. `import_data` returns a `Project` object rather than a `GenericNetwork` in all cases
5. `OpenpnmIO` is removed in favor of `Pickle`
6. the `Pickle` class saves files as `.pkl` rather than `.pnm` which is reserved for our new file format
7. Most args now have a default value to accommodate (2)
8. projects now use the proj prefix instead of sim
9. save/load_project on Workspace and Project outputs to a PNM file, which is an HDF5 file, no longer uses pickle
10. save/load_workspace on Workspace outputs a zip file containing PNM files...no longer using pickle
11. Geometry classes no longer default to all pores/throats if none are given


jgostick:
Fixes #1638 
Fixes #1637 
Fixes #1619

jgostick:
@ma-sadeghi this is ready to merge.  I implemented all the changes we discussed yesterday, I think.  I may need a bit of polishing, but we can put it on dev first, then go from there.

PR #1641, opened 2020-09-13
Using uuids on objects

### LINKED ISSUE #1594: Objects should have a more robust identifer other than name
``obj.name`` is handy and user friendly, but it is actually a bit fragile, and if for some reason the object loses is ``project`` the whole system breaks.  I think the back end should be re-factored to look for uuids instead of names.  These never change and are never in conflict. 

Given the above plan it would also be possible to keep a handle for every single object in the ``Workspace`` , indexed by uuid.  This would be a master list of objects, and could be used to query global info like which project they belong to, etc.  

Basically, I'm finding that the Workspace/Project system needs work.  :-(

### LINKED ISSUE #1618: Add name to settings, change name attr to property method
It would be super handy if the name of the object was stored in its settings, as well as any other attributes for that matter, so that we can more easily store all of an objects meta-data.  This is relevant to the new file format.

jgostick:
* Adding a uuid to each object for more robust and universal object recognition
* Improving the mechanism for assigning names to objects, and enhancing the robustness of some backend methods
* Completely new grid class, derived from a new module I created called Tableist


jgostick:
Unfortunately, I mixed a few changes into this PR, so the change log entry will be a pain.  Something like:

"Added name to settings, added uuid to all objects (stored as a hidden setting), added a new grid class called Tableist to better handle the project grid."

jgostick:
BTW, this Tablist class could/should probably be it's own package.  There are quite a few 'table' package on PyPI already, but none did quite what I wanted.

PR #1645, opened 2020-09-14
RelPerm: Fix the arrays passing to the settings 

Ni2M:
The settings arguments for arrays should be a dict instead of passing the array directly.


Ni2M:
I misunderstood a part of the code. The problem is fixed in another branch by Dr Gostick.

PR #1651, opened 2020-09-20
Speedup neighbor throat lookups

### LINKED ISSUE #1649: interpolate_data is stupidly slow for pore data
Luckily it never gets called, because we'd still be waiting for the first result to arrive.  It loops over each pore, then loops over each throat.  This can be done using numpy's accumulate functionality, along with the conns array.  This can be used to put throat properties into a pore sequentially, summing the result instead of over writing.

jgostick:
Fixes #1649


jgostick:
The interpolate_data method is now 100x faster!  I think this is ready to be merged.  There is a chance that some methods elsewhere could benefit from this trick, but we can fix those later.  The important thing is that the ``from_neighbor_throats`` method has a template for how to inspect neighbor throats in a vectorized way, using the incidence matrix, analogously to the ``from_neighbor_pores`` which uses the adjacency matrix.

jgostick:
To deal with Nans, create a dummy data array and put a suitable value in nan locations so they don't influence the result.

Issue #1575, opened 2020-07-13
Interchange front-back and left-right in accordance with the right hand rule


ma-sadeghi:
# Right hand rule (Blender manual)
<img src="https://user-images.githubusercontent.com/14086031/87323157-67c52f00-c4fc-11ea-9ebc-d06fa9b0fa5b.png" width=30%>

# Graph Algorithms and Applications I (textbook)
<img src="https://user-images.githubusercontent.com/14086031/87327567-36e7f880-c502-11ea-90c5-7e14660756a3.png" width=80%>

jgostick:
how does this compare to numpy indexing and to paraview?

ma-sadeghi:
## Create `Cubic` network from `numpy` array
```python
import numpy as np
import openpnm as op

im = np.ones((4, 5, 6))
net = op.network.CubicTemplate(im)
op.io.VTK.save(net, filename="net")
```
## Visualize in Paraview
<img src="https://user-images.githubusercontent.com/14086031/87327047-8ed22f80-c501-11ea-9cd4-8b6998841fbe.png" width=70%>

## Interchange left-right w/ front-back
```python
import numpy as np
import openpnm as op

im = np.ones((4, 5, 6))
net = op.network.CubicTemplate(im)

net["pore.left"], net["pore.front"] = net["pore.front"], net["pore.left"]
net["pore.right"], net["pore.back"] = net["pore.back"], net["pore.right"]

op.io.VTK.save(net, filename="net")
```
## Visualize in Paraview
<img src="https://user-images.githubusercontent.com/14086031/87328000-d6a58680-c502-11ea-931a-792a908fc375.png" width=70%>

Ni2M:
OpenPNM: 
YZ front,back  deltaX
XZ left, right  deltaY
XY top,bottom  deltaZ
It should be:
YZ left, right  deltaX
XZ front, back deltaY
XY top, bottom deltaZ
working on a PR.

Ni2M:
It is fixed in the PR #1660

PR #1660, opened 2020-09-25
Fix the boundary labeling problem

Ni2M:
#1575 fix the labeling problem to be consistent with the right hand rule.


Ni2M:
It's breaking the examples and tests. I'll look into them to fix it.

Ni2M:
The changes in labeling works. The test files mostly fail.
left
right
front
back
top
bottom

jgostick:
Fixes #1575

Ni2M:
The only Problem left is the pn.check_network_health output in [example tutorial advanced](https://github.com/PMEAL/OpenPNM/blob/dev/examples/tutorials/Intro%20to%20OpenPNM%20-%20Advanced.ipynb). When I run it locally on dev and on this branch, it gives me the same answer, but the ci test gives a different answer.

ma-sadeghi:
> The only Problem left is the pn.check_network_health output in [example tutorial advanced](https://github.com/PMEAL/OpenPNM/blob/dev/examples/tutorials/Intro%20to%20OpenPNM%20-%20Advanced.ipynb). When I run it locally on dev and on this branch, it gives me the same answer, but the ci test gives a different answer.

How much of a difference are we talking about? Is it like a numerical error?

Ni2M:
> 
> 
> > The only Problem left is the pn.check_network_health output in [example tutorial advanced](https://github.com/PMEAL/OpenPNM/blob/dev/examples/tutorials/Intro%20to%20OpenPNM%20-%20Advanced.ipynb). When I run it locally on dev and on this branch, it gives me the same answer, but the ci test gives a different answer.
> 
> How much of a difference are we talking about? Is it like a numerical error?

Ni2M:
@ma-sadeghi The online results generates the same array, but without the data type (dtype=int) whereas on my local computer both branches generate the dtype as well when printing the arrays. I can ignore the output check for that cell in .ipynb file.

Ni2M:
I think this branch is ready to merge, if the code factor is acceptable.

PR #1666, opened 2020-09-29
Add total_rate and rates arguments to set_rate_BC method

### LINKED ISSUE #1518: set_rate_BC accepts values as an argument
It might be more logical if this accepted ``rates`` , so:

``set_value_BC`` accepts ``values``

``set_rate_BC`` accepts ``rates``

Obviously this is very minor.  This could be done in a backward compatible way by accepted both arguments.

mkaguer:
Fixes #1518


openpnm/algorithms/GenericTransport.py
```diff
@@ -311,8 +313,22 @@ def set_rate_BC(self, pores, values, mode='merge'):
...
         ``settings``, e.g. ``alg.settings['quantity'] = 'pore.pressure'``.
         """
+        # support 'values' keyword
+        if 'values' in kwargs.keys():
+            rates = kwargs.pop("values")
+            warnings.warn("'values' has been deprecated, use 'rates' instead.",
+                          DeprecationWarning)
+        # handle total_rate feature
+        if total_rate is not None:
+            if np.array(total_rate).size != 1:
```
ma-sadeghi: @mkaguer Maybe use `np.size(total_rate) != 1` instead

ma-sadeghi: @mkaguer Also, add a unit test to cover the edge case.

openpnm/algorithms/GenericTransport.py
```diff
@@ -320,7 +320,7 @@ def set_rate_BC(self, pores, rates=None, total_rate=None, mode='merge',
                           DeprecationWarning)
         # handle total_rate feature
         if total_rate is not None:
-            if np.array(total_rate).size != 1:
+            if np.size(total_rate) != 1:
```
jgostick: numpy has an ``isscalar`` function for this check.

openpnm/algorithms/GenericTransport.py
```diff
@@ -311,8 +313,22 @@ def set_rate_BC(self, pores, values, mode='merge'):
...
         ``settings``, e.g. ``alg.settings['quantity'] = 'pore.pressure'``.
         """
+        # support 'values' keyword
+        if 'values' in kwargs.keys():
+            rates = kwargs.pop("values")
+            warnings.warn("'values' has been deprecated, use 'rates' instead.",
+                          DeprecationWarning)
+        # handle total_rate feature
+        if total_rate is not None:
+            if np.isscalar(total_rate) is False:
```
ma-sadeghi: @mkaguer use `if not np.isscalar(total_rate):` for consistency.

openpnm/algorithms/GenericTransport.py
```diff
@@ -311,8 +313,22 @@ def set_rate_BC(self, pores, values, mode='merge'):
...
         ``settings``, e.g. ``alg.settings['quantity'] = 'pore.pressure'``.
         """
+        # support 'values' keyword
+        if 'values' in kwargs.keys():
+            rates = kwargs.pop("values")
+            warnings.warn("'values' has been deprecated, use 'rates' instead.",
+                          DeprecationWarning)
+        # handle total_rate feature
+        if total_rate is not None:
+            if np.isscalar(total_rate) is not True:
```
ma-sadeghi: I think it should be `if not np.isscalar(total_rate):`. It's so-called more "Pythonic" since `isscalar` already has the verb "is", and maybe more importantly since `if bool:` already evaluates to True/False in Python, so no need to explicitly compare, e.g. `if bool is True:`, for instance.

I know it's probably a bit annoying, but that's the recommended approach (you can read up [here](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html) for a complete style guide if you'd like)

mkaguer: Hey @ma-sadeghi no problem. I changed it. If Jeff is happy, all that's left is to clear the codecov error, which I will need you to do.

jgostick: code cov is fine, it's the codefactor thing that is a pain.

jgostick:
This looks ready.  Nice work @mkaguer

ma-sadeghi:
@jgostick @mkaguer I just made some small changes to make CodeFactor happy! There are two issues left, which are basically FIXME tags, which are notes to self so I can take care of them later, so it's good that they show up (or at least that's how I think).

Anyway, let's wait for the tests and merge this. Good job @mkaguer.

ma-sadeghi:
@jgostick Actually, I just told CodeFactor not to bother us anymore with FIXME tags! So, it's all good I guess!

PR #1667, opened 2020-09-30
Add set_outflow_BC to NernstPlanck algorithms

### LINKED ISSUE #1657: Add outflow_BC to PNP

magnaou:
Fixes #1657 

- added outflow_bc to NP algs
- fixed the remove_BC in algs
- added an example
- added a unit test
- minor changes to the AdvectionDiffusionTest


mkaguer:
Hey @magnaou. I was testing out your branch and I don't think your set_outflow_BC is availabel for TransientNersntPlanck simulations. Could you check that?

magnaou:
@mkaguer will check! Thanks for reporting

ma-sadeghi:
@magnaou I took care of the CodeFactor errors (or better say, I made CodeFactor to ignore them for this PR!).

jgostick:
Before we merge this one we need to make one additional change.  It's be better if we could instantiate all these algs without specifying the phase and the ions, and allow them to be specified afterwards via the settings.

p-thomson:
Hi guys, sorry to jump on this thread but I think it is linked to what I am looking to do. I am using the TransientFickianDiffusion simulation and would like to use the set_outflow_BC, however I do not think it is currently possible?

Many thanks,

Paul

jgostick:
Hi @p-thomson...should probably open an issue for this.  We use the issue tracker as a help forum.  But outflow bc only makes sense if there is flow, as in advection diffusion.  Open an issue and explain what sort of bc you're looking for.

mkaguer:
Hey @magnaou. I was just thinking that it might be best to remove both diffusive and migrative contributions from the conductance. That leaves only flow! For my cdi problem I need dc/dx=0 and dphi/dx=0. Just thought I would add that here for you to consider.

ma-sadeghi:
Hi @magnaou, the test you added for `test_outflow_BC` is not passing. How can we ensure that `set_outflow_BC` is doing what it's supposed to do?

Like for example in `AdvectionDiffusion`, a good case to compare is to have flow from left to right, and impose Dirichlet BC on the left face, `c = 1.5`, and outflow BC on the right, and expect that the concentration of the right pores be `c = 1.5` since there's no consumption (actually this is how we test it in our unit tests).

My question is: how can we perform such simulation so we're sure the function works as expected.

Issue #1671, opened 2020-10-05
Allows Physics instantiation without Phase, and ability to associate with Phase afterward


jgostick:
This is pretty subtle, but would help make the code more flexible and forgiving.  We've been doing several of these sort of relaxing of conditions lately, so this is part of a larger change.  This is stemming from the new PNM IO class, which needs instantiation and object associations to be as flexible as possible (ie. the ability to create things in opposite orders).

jgostick:
After digging into the code, I see that we already 'try' to allow this, but it doesn't work.  The project look-up breaks when no phase is given.  I will tag this as a bug.

jgostick:
We already support this, BUT it breaks if the physics has any models on it.  The init should be changed add 'freeze_models' to the settings if phase is None.

jgostick:
closed by #1674

Issue #1670, opened 2020-10-05
Specialize Exceptions


ma-sadeghi:
While it's considered best practice to use Python's default Exceptions if possible, I think adding a few of our own might be worth it. I propose adding the following Exceptions:

- `SolverError` -> `NotConvergedError`
- `MissingDependencyError`
- `NetworkTopologyError`

jgostick:
Great point...I'd been pondering this recently too.

jgostick:
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

jgostick:
I'm actually questioning the wisdom of this.  At the moment we print a descriptive message with each exception, and the python trace back is very detailed.  I am not sure that adding specialized exceptions will make things any more clear, but I am sure that they will be a huge pain in the butt to code, since you have to try/accept based on the expected exception, so this will add a lot of uncertainty to the mix.

jgostick:
@ma-sadeghi posted a talk to the teams channel a few weeks ago, in which the author said that subclassing exceptions was terrible idea since it led to class creep.  I'm inclinded to agree.  We should stick to the available exceptions, and just add useful messages.

PR #1674, opened 2020-10-05
Instantiate Physics with no Phase

jgostick:
I have updated the ModelsMixin class to catch all ``Exceptions`` and set model to deferred instead throwing an exception.  This means that instantiating a class when things are not quite setup properly is still possible, allowing the user to manually fix things afterwards, like create associations between Physics and Phases.  This change will actually help with classes where models are added but cannot be run.  Before this change we only caught a KeyError exception, if a dependent property was not found.


jgostick:
We need to have a group discussion about this.  The way I've implemented things, all 'exceptions' that occur when adding a model are caught, the model is added but not run, and the regen_mode is set to 'deferred'.  My thought was that an error message is issued, which includes the exception message, so the use can then go fix the problem.  This way the code can still run and be 'responsive', rather than halting fatally...but I wonder if it's a bad idea to have a program that appears to be running but is actually in a 'broken' state?

jgostick:
Interestingly, we could solve this problem by implementing the idea that @ma-sadeghi mentioned this morning in #1670.  If we had more specialized exceptions, then we could catch some of them and raise a warning, but let others fail?  I guess that begs the points, why have an exception if you don't plan to halt execution if it's found.

ma-sadeghi:
I think letting the user write code that doesn't run might be dangerous, as they might go too far. Basically, I think fixing errors as soon as they appear is much easier than fixing an entire program that's broken on too many fronts. 

Another advantage of not letting users write broken code is that they can "learn" how to write proper OpenPNM code as they write small snippets that occasionally fail to run. Anyways, just my opinion.

jgostick:
Investigate the 'freeze_models' setting to see if that helps.

jgostick:
Fixes #1671

jgostick:
This version sets ``phys.settings['freeze_models'] = True`` if ``phase=None``.  This solves the problem perfectly, except the user now needs to 'unfreeze' the models.  I have added a warning message to the `regenerate_props` method to inform the user of the situation...this is the best we can do I think.  If this passes all tests, then it can be merged.  The changes were very simple in the end.

ma-sadeghi:
> This version sets `phys.settings['freeze_models'] = True` if `phase=None`. This solves the problem perfectly, except the user now needs to 'unfreeze' the models. I have added a warning message to the `regenerate_props` method to inform the user of the situation...this is the best we can do I think. If this passes all tests, then it can be merged. The changes were very simple in the end.

Can we add a helper method that "unfreezes" all models? and let the user know in the warning message to call that method when they're done.

jgostick:
We could do something like:

```
if phase is None:
    self.settings['freeze_models'] = True

# Do rest of init here:

if phase is None:
    self.settings['freeze_models'] = False

```

ma-sadeghi:
> We could do something like:
> 
> ```
> if phase is None:
>     self.settings['freeze_models'] = True
> 
> # Do rest of init here:
> 
> if phase is None:
>     self.settings['freeze_models'] = False
> ```

Sounds like a good idea. Maybe also add a comment explaining why we do this, as a note to future self.

jgostick:
Sorry, this doesn't work.  We'd have to add this at the super.__init__ level, which gets called before the subclassed stuff runs, which is where all the models are.  Unless....maybe python has a magic method that gets called 'after' a subclass is finished with it's own init?

ma-sadeghi:
> Sorry, this doesn't work. We'd have to add this at the super.**init** level, which gets called before the subclassed stuff runs, which is where all the models are. Unless....maybe python has a magic method that gets called 'after' a subclass is finished with it's own init?

Is this relevant by any chance? ([source](https://stackoverflow.com/questions/16017397/injecting-function-call-after-init-with-decorator))
```python
# define a new metaclass which overrides the "__call__" function
class NewInitCaller(type):
    def __call__(cls, *args, **kwargs):
        """Called when you call MyNewClass() """
        obj = type.__call__(cls, *args, **kwargs)
        obj.new_init()
        return obj


# then create a new class with the __metaclass__ set as our custom metaclass
class MyNewClass(object):
    __metaclass__ = NewInitCaller
    def __init__(self):
        print "Init class"
    def new_init(self):
        print "New init!!"

# when you create an instance
a = MyNewClass()
>>> Init class
>>> New init!!
```

jgostick:
Technically that is 'the' solution (or the python 3 version), but I feel like this would make subclassing the Generic classes very complicated.  A user would have to pass in the metaclass and call it.  Might as well just add ``if phase is None: self.settings['freeze_models']`` to each new subclass.  

I actually spent 30 mins skimming through old rejected PEPs to see if anyone has proposed a ``__post__`` method that get called after each ``__init__``.  I couldn't really tell from the pep titles, but it looks like not.

ma-sadeghi:
> Technically that is 'the' solution (or the python 3 version), but I feel like this would make subclassing the Generic classes very complicated. A user would have to pass in the metaclass and call it. Might as well just add `if phase is None: self.settings['freeze_models']` to each new subclass.
> 
> I actually spent 30 mins skimming through old rejected PEPs to see if anyone has proposed a `__post__` method that get called after each `__init__`. I couldn't really tell from the pep titles, but it looks like not.

Do you think it would be a bad idea if we manually "unfreeze" the models when the user calls `alg.run()`?

PR #1683, opened 2020-10-07
Fix github actions deprecations

ma-sadeghi:


ma-sadeghi:
@jgostick Can you please check the one-line change I made to Workspace.py?

openpnm/utils/Workspace.py
```diff
@@ -53,7 +53,8 @@ def __setitem__(self, name, project):
         if name is None:
             name = self._gen_name()
         if name in self.keys():
-            raise Exception("A project named " + name + " already exists")
+            if id(self.__instance__[name]) != id(project):
```
jgostick: I don't 'get' this...what was wrong with the original way, and how does this help?

ma-sadeghi: Say you have a project called `proj`. Normally, if you want to change the object name for some reason, you change it to a different name than the original. But, you should still be able to run the following code:
```python
proj.name = proj.name
```
without any errors, although the above code doesn't do anything. I encountered this when I wanted to make sure that the project name is always a particular name, but in some cases, the project name was already as I hoped.

PR #1689, opened 2020-10-09
Ensure object names are unique within a project

### LINKED ISSUE #1687: Object names must be unique within a project
Currently, this is not enforced.

### LINKED ISSUE #1688: Renaming objects to their current name should be allowed
Currently, an exception is thrown complaining that another object already exists with the same name.

ma-sadeghi:
Fixes #1687 
Fixes #1688

### DIFF
--- a/openpnm/utils/Workspace.py
+++ b/openpnm/utils/Workspace.py
@@ -53,8 +53,9 @@ class Workspace(dict):
         if name is None:
             name = self._gen_name()
         if name in self.keys():
-            if id(self.__instance__[name]) != id(project):
-                raise Exception("A project named " + name + " already exists")
+            # To ensures renaming a Project to its current name is allowed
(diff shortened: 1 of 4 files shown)


ma-sadeghi:
@jgostick Can you please review this PR?

openpnm/utils/Workspace.py
```diff
@@ -53,7 +53,9 @@ def __setitem__(self, name, project):
         if name is None:
             name = self._gen_name()
         if name in self.keys():
-            raise Exception("A project named " + name + " already exists")
+            # To ensures renaming a Project to its current name is allowed
```
jgostick: I still don't understand these 3 lines.  I've never used the __instance__ magic before.  Also I don't understand why if the object is NOT in the project we raise an exception.  I guess if it works, then that is all that matters.

PR #1691, opened 2020-10-09
using __new__ to recreate all classes upon load

jgostick:
This very minor change to the PNM.load_project method should allow us to relieve the requirement of class instantiate with no args.  If the tests pass, I will then begin to revert the network classes back to the old formats.


jgostick:
I've realized that this method will not work with certain classes, particularly the voronoi related networks.  These classes save the voronoi and delaunay tessellations on themselves as hidden attributes.  

- The tessellations are objects and not serializable
- They cannot be stored in the settings dict
- They cannot be generated after the fact

The networks will still work (as illustrated by the fact that all tests pass), but should the user go looking for the vor or tri attributes, they'll get errors.  

I'm inclined to accept this for now and make a note to fix it later.

jgostick:
@ma-sadeghi this is as ready as I can reasonably make it, though still not perfect.  I'd like to merge this soon, so we can then merge examples_galore (#1513).  Let me know when you have a few minutes and we can go over it on teams to discuss the important bits.

PR #1701, opened 2020-10-17
Salome export

magnaou:


jgostick:
I was mistaken...it seems you have already merge Salome into dev...I must've been looking in the wrong branch!  Sorry. I guess the plus side is that you added some tests :-)

Issue #1407, opened 2020-02-22
Allow 're-assignment' of pores when creating Geometry and Physics?


jgostick:
at the moment, if you assign all 100 pores to a geometry, then try to create a second geometry, we throw an error.  Maybe it would be convenient to allow this.  For instance, if I'm importing a network, then I use the `Imported` geometry class, all the pores are assigned to it.  But what if I subsequently want to create a boundary geometry and assign pores 1, 2 and 3?  It may be handy to allow this:

``` python
geo1 = openpnm.geometry.Imported(network=pn)
geo2 = openpnm.geometry.Boundary(network=pn, pores=[1, 2, 3])
```

Interestingly, we used to offer `set_locations` and `remove_locations` methods, but have removed them.  So there is not actually a way to do this.  It used to work like this:

``` python
geo1 = openpnm.geometry.Imported(network=pn)
geo2 = openpnm.geometry.Boundary(network=pn, pores=None, throats=None)
geo1.remove_locations(pores=[1, 2, 3])
geo2.set_locations(pores=[1, 2, 3])
```

jgostick:
Now that I'm playing with it a bit, here is a proposal:

This should be handled by the `Project`.  We should add a function like `reassign_locations`, that shifts all info from one object to another, including pore scale models, properties, labels, etc. Like this:

``` python
proj.reassign_locations(source=geo1, target=geo1, pores=[1, 2, 3], throats=[0, 5, 8], mode='all')
```

`mode` could also be a list like ['labels', 'props', 'models']

jgostick:
This was addressed in PR #1705

PR #1705, opened 2020-10-17
Expose set locations

jgostick:
This PR focuses on providing the user with more power to change physics and geometry associations.  

The hidden ``_set_locations`` method on the Subdomain class now has two public wrappers on the GenericGeometry class, ``add_locations`` and ``drop_locations``.  These both call the ``_set_locations`` method, and do a bit of extra clean-up.  Crucially, these methods will also change the locations of any associated physics objects.  If I remember correctly, we had hidden these methods because we didn't want users to start fiddling with locations on a geometry, since this would break the associated physics.  So we made it difficult to change these associations after the objects were made.  This new approach addresses this, since it keeps things correctly connected.

This also adds ``set_phase`` and ``set_geometry`` method to the GenericPhysics class.  Changing phase is straightfoward.  Changing geometry effectively changes which pores and throats the physics is associated with.  Each of the methods has several modes, including 'add', 'drop', and 'swap'.  

Unit tests were added for all new methods.

### DIFF
--- a/openpnm/core/Base.py
+++ b/openpnm/core/Base.py
@@ -395,7 +395,7 @@ class Base(dict):
     def keys(self, element=None, mode=None, deep=False):
         r"""
         This subclass works exactly like ``keys`` when no arguments are passed,
-        but optionally accepts an ``element`` and/or a ``mode``, which filters
+        but optionally accepts an ``element`` and a ``mode``, which filters
         the output to only the requested keys.
 
         The default behavior is exactly equivalent to the normal ``keys``
--- a/openpnm/core/Subdomain.py
+++ b/openpnm/core/Subdomain.py
@@ -10,20 +10,6 @@ class Subdomain(Base):
 
     Notes
     -----
-    The following table list the two methods added to Base by this subclass.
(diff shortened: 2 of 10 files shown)


jgostick:
This PR will need a bit of discussion.  I basically doubles-down on the use of Geometry objects to define where Physics object apply.  In principle, a single physics object *can* be applied to an entire domain; however, lots of 'look-ups' depend on a 1-to-1 correspondence between geometries and physics.  Note that this current PR is 100% backwards compatible with our tests.  We could, alternatively, have reverted back to the style of version 1, where physics are assigned by specific pores and throats.  I am not sure if that would have been backwards-compatible or not.  I can say that the data doesn't really care if physics have a 1-to-1 match with geometries (ie. interleave data, interpolate data, models, etc would all work).  It's just the object look-ups at the project level that have a problem.

jgostick:
Fixes #1407

jgostick:
This is now ready to go.  I fixed the bugs we encountered during yesterday's meeting and also fixed a bug in the grid.

Issue #1684, opened 2020-10-08
Workspace.load_project returns None, but maybe should return project object


ma-sadeghi:
Currently, the `load_project` method on our IO classes returns a project object, maybe `Workspace.load_project` should do the same? @jgostick was this intended?

jgostick:
That's maybe a good idea.  At the moment it just loads the project into the workspace, but does not return the handle.  That's an easy fix if you think it makes more sense.

PR #1715, opened 2020-10-23
Workspace should returns the project, rather than nothing or the network

Ni2M:
This fixes issue #1684


tests/unit/utils/WorkspaceTest.py
```diff
@@ -56,8 +56,11 @@ def test_save_and_load_project(self):
         self.ws.close_project(proj)
         assert 'test_proj' not in self.ws.keys()
         assert proj.workspace == {}
-        self.ws.load_project(filename='test_proj.pnm')
+        proj = self.ws.load_project(filename = 'test_proj.pnm')
```
ma-sadeghi: I think you should assert that `proj` is of type `Project`.

ma-sadeghi (approved):

PR #1740, opened 2020-11-06
Build documentation ourselves using GitHub Actions and host it on GitHub

ma-sadeghi:
We've been having problems with `readthedocs` building our documentation. I looked into it and it seems it has something to do with `setup.py` of one of `numba`'s dependencies. Long story short, until they fix the `setup.py`, our documentation will not be automatically built. 

With this PR, our documentation is currently hosted on GitHub: http://pmeal.github.io/OpenPNM

There's only one issue: the address is case-sensitive. I was wondering if we should lowercase the repo name for consistency with the package name.

- [x] Update front page of docs to look a bit more 'active'
- [x] Move conda install to the main installation method on front page
- [x] Update readme to point to github.io
- [x] Update setup.py


jgostick:
- [ ] Update front page of docs to look a bit more 'active'
- [ ] Move conda install to the main installation method on front page
- [ ] Update readme to point to github.io

Issue #1743, opened 2020-11-06
Ensure model parameters are serializable


jgostick:
This is done in the setting dicts, but should be done in the models dict as well, to ensure everything can be saved in a pnm file.

jgostick:
We have decided to abandon this and stop supporting 'full' saves to non-pickle formats.

Issue #1718, opened 2020-10-23
NernstPlanck related algs should only accept ion objects, not name strings


jgostick:

jgostick:
We are completely overhauling the PNP and multiphysics in general. This issue will not exist in the near future.

PR #1745, opened 2020-11-07
Nonserializable check in models

jgostick:
Simple addition to the ModelsWrapper class init to ensure all model args, except for models themselves, are serializable.


jgostick:
It seems there are 19 failed tests.  Most are due to passing in a phase object rather than a name, but there are a few that are literally designed to accept function, such as the symbol (sympy) source term models, and the generic pore/throat size models that accept a frozen scipy.stats models.  Hmmm.

jgostick:
I have fixed the symbolic source term (backwards incompatibly), and it now works as I always thought it should.  It accepts a string describing the equation, and key word arguments.  Everything else is done internally.

Now I want to fix the generic pore size model.  I think it also needs to be done backwards incompatibly.  If we change the ``func`` argument to ``func_name``, and accept a string, we can then use ``getattr(scipy.stats, func_name)`` internally.  Simple to fix, but incompatible.

jgostick:
This fixes #1743, as well as #1718

jgostick:
This one will need some discussion.  It breaks backward compatibility for several models, and could pose an major annoyance to people by blocking the passing of numpy arrays to the models.  Although, I would argue that not being able to save the work to a PNM file is a greater annoyance.

ma-sadeghi:
@jgostick Can we merge this?

PR #1772, opened 2020-11-20
Fix the use of csgraph.laplacian for building A

### LINKED ISSUE #1540: Computing conductance values using Laplacian operator is convoluted
Basically, for calculating conductance values, we use Scipy's `laplacian` operator with default arguments, i.e. `use_out_degree=False`. This is fine for symmetric problems, but explaining it becomes very convoluted when the problem is assymetric, such as mass partitioning or advection-diffusion. I researched the matter quite a bit and I think we should use `laplacian` with `use_out_degree=True`. This transition needs a thorough analysis of the code to make sure we don't break anything, although we have tons of unit tests. Anyways, this ways, explaining why we did what we did in `multiphase_diffusion` pore-scale model becomes almost obvious to someone reading the code. This issue was brought up while I was trying to explain it to @mkaguer.

Ni2M:
Fixes #1540.


ma-sadeghi:
Following the discussion with @Ni2M, I think this PR is no longer necessary.

Issue #1738, opened 2020-11-05
Pass initial condition to phase where it is used to update physics models prior to alg.run()


mkaguer:
It would help if alg.set_IC() passes initial conditions to phase object as 'pore.quantity'. Then physics models can be rerun so that upon running algorithm all physics models start at correct initial condition. 

This problem came up for me when running a nernst planck simulation where ionic_conductance depends on pore.concentration but setting initial condition did not change starting value for ionic_conductance. The work around was setting phase['pore.concentration'] = initial_condition prior to adding ionic conductance model.

This issue is related to issue #1478 in how initial guess could be written to phase object. Should initial guess be initial condition?

jgostick:
In my latest version I got the ICs and value BCs to both fill the 'pore.ic' array as well as the 'pore.quantity' array.  There is one problem though...the crank-nicolson scheme has different values.  There are hard coded values in the test, and now they've changed, so I'm guess that the previous version was not including BCs in the initial time step, and they didn't get added until the second time step.  This would corroborate some results I've seen lately.

PR #1792, opened 2020-11-26
Pass initial condition to phase object and regenerate physics models as part of set_IC()

mkaguer:
Trying to fix issue #1738 

Right now I am regenerating all physics models. I am wondering if this could cause problems. Is there a way to regenerate only the physics models that depend on `quantity`? If so, I think that would be a better way to handle this.

### DIFF
--- a/openpnm/algorithms/TransientReactiveTransport.py
+++ b/openpnm/algorithms/TransientReactiveTransport.py
@@ -161,7 +161,9 @@ class TransientReactiveTransport(ReactiveTransport):
 
     def set_IC(self, values):
         r"""
-        A method to set simulation initial conditions
+        A method to set simulation initial conditions. The values for initial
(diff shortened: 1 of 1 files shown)


ma-sadeghi:
@mkaguer There's a method on `GenericTransport` (or maybe `ReactiveTransport`) called `_update_iterative_props`. You can take a look and see the logic, maybe you can take parts of it for what you're trying to achieve.

jgostick:
Hi @mkaguer...I'm sorry, I went ahead and dove into this issue in a different PR (#1795).  I got things working now, but there is on test failing.  The transient solver with crank-nicolson scheme is producing different results.  This may mean the old results were flawed, or it may mean I didn't something wrong.  Once we get that sorted out I think it is good to go.

PR #1793, opened 2020-11-28
Changing _get_solver behavior to fall back to scipy if pardiso missing

### LINKED ISSUE #1710: Change default solver to scipy's spsolve if pypardiso is not found
Currently, the default solver is `pypardiso`, which relies on MKL, and installing MKL using `pip` seems to be not so straightforward.

jgostick:
fixes #1710

### DIFF
--- a/openpnm/algorithms/GenericTransport.py
+++ b/openpnm/algorithms/GenericTransport.py
@@ -634,11 +634,6 @@ class GenericTransport(GenericAlgorithm):
         """
         # SciPy
         if self.settings['solver_family'] == 'scipy':
-            if self.Np > 8000:
-                logger.warning('The default scipy solver will be very slow '
-                               + 'on a network this large. We strongly '
-                               + 'recommend using pardiso.')
-
             def solver(A, b, atol=None, rtol=None, max_it=None, x0=None):
                 r"""
                 Wrapper method for scipy sparse linear solvers.
--- a/openpnm/algorithms/GenericTransport.py
+++ b/openpnm/algorithms/GenericTransport.py
@@ -669,11 +669,19 @@ class GenericTransport(GenericAlgorithm):
                 return x
         # PyPardiso
         elif self.settings['solver_family'] == 'pypardiso':
+            try:
+                import pypardiso
+            except ModuleNotFoundError:
+                logger.critical("Pardiso not found, reverting to much "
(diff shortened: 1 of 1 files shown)


jgostick:
This is hard to test since it literally requires an env without pardiso, which is installed on the tests envs.  @ma-sadeghi , can you take a look at what I did to make sure it follows the logic of how you have things setup?

ma-sadeghi:
@jgostick I thought we decided to throw an Exception with instructions on how to install pypardiso, rather than falling back to spsolve.

jgostick:
Hmm, I guess it wasn't clear.  Anyway, I think @Zohaib-Atiq's point about making sure that OpenPNM always 'works' is a good one.  The warning message I added it at the 'critical' level so doesn't get silenced, it makes it clear that revert to scipy is 'much slower', and gives instructions for installing pardiso.

ma-sadeghi:
Right, it wasn't clear. Anyway, I don't want to sound arrogant, but I still think we shouldn't allow that. Installing `pypardiso` takes a few seconds, and our readme is clear on using `conda` for installing `OpenPNM`, which comes with `pypardiso` preinstalled.

Just to remind ourselves how slow umfpack-less `spsolve` can be, here's a quick benchmark:

| Number of pores  | 30^3 | 40^3 | 50^3  | 60^3  |
|------------------|------|------|-------|-------|
| pardiso4py (s)   | 0.15 | 0.7  | 2.4   | 5.6   |
| spsolve (s)      | 3.6  | 25.3 | 140.0 | 551.0 |
| Slow-down factor | 25.1 | 36.2 | 58.1  | 98.2  |

The main reason I think we shouldn't allow scipy's `spsolve` is that it's not _just slow_, it virtually makes it impossible to work with `OpenPNM` past 40-50ish cubed pores, and also that installing `pardiso4py` is as reliable as other dependencies we have (since conda-forge distributes cross-platform binaries).

That said, if you guys think the fall back is necessary, it's all good.

jgostick:
Having well founded opinions is not arrogant. Here's a compromise: Check the network size, and if Np > 10,000 or so, then throw an exception, otherwise just issue the warning.

ma-sadeghi:
I think that's a good idea. Also, the 10,000-pore threshold seems spot on as `spsolve` is still within the comfort zone 😄 (0.45 s for ~10,000 pores).

PR #1794, opened 2020-11-28
Prevent overwriting on outflow bc

### LINKED ISSUE #1747: Outflow BC and other BC types must be prevented to coexist in a given set of pores

jgostick:
fixes #1747

### DIFF
--- a/openpnm/algorithms/AdvectionDiffusion.py
+++ b/openpnm/algorithms/AdvectionDiffusion.py
@@ -172,6 +172,7 @@ class AdvectionDiffusion(ReactiveTransport):
         self.A.setdiag(diag)
 
     def _set_BC(self, pores, bctype, bcvalues=None, mode='merge'):
+        pores = self._parse_indices(pores)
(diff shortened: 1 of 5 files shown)


jgostick:
This same logic should be applied to PNP algs too.  Maybe that should be opened as a separate issue?

jgostick:
I have beefed up the _set_bc method to prevent overwriting of all bcs, not just outflow, depending on mode.  I added a new mode called insert which does allow overwriting of existing bcs.  This is apparently backward incompatible.   I had to change the mode of ONE test to 'insert' instead of the default 'merge'.  Merge used to allow over writing despite the description in the docstring that it does NOT.  Sooo, technically this is a bug fix, but it might break peoples code since the default behavior (merge) works differently.  We could change the default mode to 'insert' which would should maintain the old behavior...but this also seems somewhat backward incompatible.

jgostick:
I rolled back the 'insert' mode, since that is how 'merge' was acting (even though I'm not sure that was the intent...the docstring was not super clear actually).

PR #1795, opened 2020-11-29
Setting quantity on phase when IC is called

jgostick:
Closing #1738


mkaguer:
Hi @jgostick. I reviewed the code you merged. It looks good. However, I do not think it will work for Nernst Planck simulations. Nernst Planck uses it's own run method so it does not access the run method in TransientReactiveTransport where you call _overwrite_ICs_with_value_BCs(). I think this is a sign, however, that the nernst planck solvers need some refactoring or something like that. Why not have each of the algorithms in NP call their own run method? That will be a future improvement separate from this I think.

PR #1800, opened 2020-12-01
Add mandatory args (either pores or throats) to geometry

### LINKED ISSUE #1784: Raise exception if no pores/throats are passed to Geometry
This is similar to #1709

ma-sadeghi:
Fixes #1784

### DIFF
--- a/openpnm/geometry/GenericGeometry.py
+++ b/openpnm/geometry/GenericGeometry.py
@@ -68,6 +68,10 @@ class GenericGeometry(Subdomain, ModelsMixin):
 
     def __init__(self, network=None, project=None, pores=[], throats=[],
                  settings={}, **kwargs):
+        # Make sure at least one of [pores, throats] is not empty
+        if len(pores) == 0 and len(throats) == 0:
+            raise Exception("'pores' and 'throats' can't be both empty.")
+
         # Define some default settings
         self.settings.update({'prefix': 'geo'})
         # Overwrite with user supplied settings, if any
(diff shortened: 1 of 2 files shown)


ma-sadeghi:
@jgostick it seems fixing this was not as trivial as I thought. The issue is that `Project._new_object` instantiates geometry with empty `pores` and `throats`:
```python
    def _new_object(self, objtype, name=None):
        r"""
        """
        if objtype.startswith('net'):
            obj = openpnm.network.GenericNetwork(project=self, name=name)
        elif objtype.startswith('geo'):
            obj = openpnm.geometry.GenericGeometry(project=self, name=name,
                                                   pores=[], throats=[])
        elif objtype.startswith('pha'):
            obj = openpnm.phases.GenericPhase(project=self, name=name)
        elif objtype.startswith('phy'):
            obj = openpnm.physics.GenericPhysics(project=self, name=name)
        elif objtype.startswith('alg'):
            obj = openpnm.algorithms.GenericAlgorithm(project=self, name=name)
        else:
            obj = openpnm.core.Base(project=self, name=name)
        return obj
```
There are two alternative solutions that come to my mind:

1. Switch back to assuming _all_ pores/throats if they're not specified (does this break the PNM format?)
2. Do a runtime geometry sanity check to look for unassigned pores/throats.

What do you think?

jgostick:
we have addressed this issue by adding calls to validate data health in the run methods, so a useful exception is raises instead of matrix is singular

PR #1801, opened 2020-12-01
Enhance show_hist layout

ma-sadeghi:
# Before
aadf

# After
aa


jgostick:
wow, I love it.  I had added that function as a quick visualization and didn't give it enough love.  Nice work.

PR #1806, opened 2020-12-04
Call overwrite_ICs_with_value_BCs method in TransientNernstPlanckMultiphysicsSolver

mkaguer:
Here is what I did in this PR:
- I added a _update_iterative_props() method call in the new overwrite_ICs_with_value_BCs() method. I did this because, otherwise, conductance models have no chance to be updated before the A matrix is built in run. Which was the point - to have all conductance values starting at their initial condition (or value BC). Right?
- I added an overwrite_ICs_with_value_BCs() method call to TransientNernstPlanckMultiphysicsSolver run method. In practice, I think each algorithm should be calling their own run method which already has overwrite_ICs_with_value_BCs() incorporated but that's not how the solver currently works so I had to do this.
- Lastly, my exact issue was with regards to ionic conductance not starting at initial condition as it depends on concentration. Relying on iterative_props to update physics models (as we did) does not fix my problem as ionic conductance is apparently not an iterative prop. So I've sort of just figured that it is still best for me to define my initial conditions manually as phase['quantity'] before adding ionic conductance model to physics.

### DIFF
--- a/openpnm/algorithms/TransientNernstPlanckMultiphysicsSolver.py
+++ b/openpnm/algorithms/TransientNernstPlanckMultiphysicsSolver.py
@@ -103,7 +103,11 @@ class TransientNernstPlanckMultiphysicsSolver(NernstPlanckMultiphysicsSolver):
                     alg.set_IC(phase[alg.settings['quantity']])
                 except KeyError:
                     alg.set_IC(0)
-
+                    
+        # Start algorithm physics with correct ICs and value BCs
+        for alg in np.flip(algs): # flip to update e_algs first
+            alg._overwrite_ICs_with_value_BCs() 
+        
         for e in e_alg:
             # Save A matrix of the steady sys of eqs (WITHOUT BCs applied)
             e._build_A()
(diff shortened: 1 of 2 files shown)


ma-sadeghi (commented):
@mkaguer what do you mean by e_algs?

ma-sadeghi:
> ionic conductance is apparently not an iterative prop.

Why do you need it to be an iterative prop? Does it change?

mkaguer:
@ma-sadeghi by e_algs I mean the nernst planck algorithms excluding the charge conservation algorithm. The e_algs algorithms need updating first because the ionic conductance in the charge conservation depends on concentration. So update e_algs to have the pore.concentration property at correct ICs/BCs then update p_algs so ionic conductance is rerun with correct ICs/BCs for pore concentration. 

Yes, ionic conductance does change. It depends on concentration.

ma-sadeghi:
@mkaguer I see. Is there a way you can refactor the charge conservation model so `pore.concentration` is an explicit dependency? In my opinion, we should avoid these hidden dependencies.

mkaguer:
@ma-sadeghi I think pore.concentration already is an explicit dependency. Check out [line 23](https://github.com/PMEAL/OpenPNM/blob/b80ad0fb5862005f3a54857d24b8e1e83d531b33/openpnm/models/physics/ionic_conductance.py#L23) in ionic coductance model. Could it be a problem with how the model is not explicitly dependent on pore.concentration.ion?

jgostick:
This solution is not general enough, and will be fixed a different way when we renovate the PNP algorithms, so closing.

PR #1813, opened 2020-12-22
Knudsen diffusion in Multiphase class

Ni2M:
We have a model for Knudsen diffusion when the void space is not occupied by liquid phase. (Note that literatures are considering no dissolution of gas in the liquid (impermeable boundary of gas/liq). To consider the liquid phase and dissolution of the gas in the liquid we can integrate the mixed_diffusion and multi_phase diffusion methods. There are some challenges to address. A preliminary change is applied to the diffusive conductance models.

### DIFF
--- a/openpnm/models/physics/diffusive_conductance.py
+++ b/openpnm/models/physics/diffusive_conductance.py
@@ -11,7 +11,10 @@ __all__ = [
     "taylor_aris_diffusion",
     "classic_ordinary_diffusion",
     "multiphase_diffusion",
-    "mixed_multiphase_diffusion"
+    "mixed_multiphase_diffusion",
+    "mixed_multiphase_diffusion_MW",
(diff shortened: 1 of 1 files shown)


Ni2M:
approach 1: Considering the presence of liquid with diffusion coefficient (my interpretation may be wrong)
image
approach 2: labeling the gas-filled pores and only apply Knudsen diffusion model on those regions
approach 3: considering the partial saturation and correction on pore size (not applicable for the multiphase class/dissolution)

Ni2M:
The problem can be solved by using two different physics. Instead of working on mixed-diffusion model, change the diffusivity of the phase in air-filled pores. Calling the multiphase_diffusive_conductance autonatically gets the diffusivity of the occupant phase.

Note: we won't consider the presence of water and reduction of available pore size, because there is the air dissolution (no impermeable boundary between air and water).
I will prepare an example related and post on the website.

PR #1821, opened 2021-01-23
Adding paraview submodule in io

jgostick:


ma-sadeghi (commented):
Great work. Just a quick question: why is the `export_data` method not part of a class (e.g. `ParaView`) like other IO classes?

jgostick:
I think this was based on @ma-sadeghi's suggestion that the io functions don't really gain anything by being classes.  There is very little inheritance happening, and we actually have the method exposed as classmethods, so they appear as functions.  I think I told @hfathian to do it this way in preparation for possibly doing this for all io classes in the future.

ma-sadeghi:
I spent some time debugging the CI failures, but no luck. In the end, I removed `paraview` from dependencies and commented its unit tests, and everything seems to work out just fine. CI machines don't have Spyder installed, so the conflict must be coming from somewhere else. `paraview` seems to be a burden at the moment. Not sure how to proceed.

ma-sadeghi:
**Update**: I continued working on debugging this PR. I tried different things on a toy branch (test_pv_deps), and it turned out that there was something wrong with the way `ParaView` module was imported. Anyways, I couldn't fix it, so I changed it to a Python class, like the other IO, and everything just works fine again!

(I know we're going to eventually convert all IO classes to just methods, but I couldn't find an alternative solution)

ma-sadeghi (approved):

PR #1825, opened 2021-02-01
Continuum geometry

jgostick:
The idea is that we can have 'continuum' nodes in a network, with effective properties.  The 'trick' is that we wanted the cells to be anisotropic.  The length and area calculations need to account for this.


ma-sadeghi:
Hey @jgostick @mkaguer, is this PR ready?

PR #1833, opened 2021-02-02
Fixing the is_fully_connected check in alg.run

jgostick:


ma-sadeghi (approved):
Great. I just did a minor refactoring.

PR #1840, opened 2021-02-07
Add statoil export data

jgostick:
This PR adds an ``export_data`` method to the Statoil IO class.  It outputs the 4 "dat" files, and is meant to allow the use of the ``pnflow`` code of the ICL group.  It can also be useful for importing into other software that uses the Statoil format.


jgostick:
I'm going to merge this even though the windows build is failing.  The tests all pass, so the problem is on the windows CI, not us

Issue #1796, opened 2020-11-30
ReactiveTransientTransport breaks when t_final == t_initial


ma-sadeghi:

mkaguer:
I looked into this some more. The problem was with [line 378](https://github.com/PMEAL/OpenPNM/blob/11530fa4ce00a80700429710bad3b26bc540d600/openpnm/algorithms/TransientReactiveTransport.py#L378)

when t_initial == t_final, `np.arange(t+dt, tf+dt, dt)` results in an empty array so when time is defined as `for time in np.arange(t+dt, tf+dt, dt)` the variable time never gets set and the following error occurs:

UnboundLocalError: local variable 'time' referenced before assignment

So I initialized time = t at the start of time marching. That is the simplest solution I could think of. @ma-sadeghi would you have a better suggestion? I thought maybe there is a way we could control the endpoints in np.arange to include t_initial. But even if you did have better control as in np.linspace and can include t_initial you don't want to call _t_run_reactive method for t = 0. We already have solution at t = 0 which is the initial conditions. Is there a use case for t_initial == t_final? Not sure why you would need to solve for the initial time when you already know it.

ma-sadeghi:
@mkaguer Thanks for looking into this. Now that I revisited `TransientReactiveTransport`, I think the `run` method is doing more than it should. I don't think the `run` method should check for whether or not the steady state condition has been reached. That's unexpected behavior, since the user asks for the solution until `t_max`, but the solver terminates once steady state is reached. 

Maybe that should be a separate method or mode. @jgostick What do you think?

mkaguer:
@ma-sadeghi that's a fair point. Another thing to consider is that the user does have control over what is considered steady. For example, the user could use really small tolerance such as 1e-32 to force solver to reach tfinal.

I am tempted to suggest that the solver should throw an error when tinitial equals tfinal. Because why would you ever want that? Thoughts?

ma-sadeghi:
That's also fine. As a general rule, though, we should avoid adhoc treatment, so the code just works. I'd propose get rid of the `if` clause (which checks whether steady has reached) in here:

https://github.com/PMEAL/OpenPNM/blob/11530fa4ce00a80700429710bad3b26bc540d600/openpnm/algorithms/TransientReactiveTransport.py#L348-L349

and similarly in here:
[...]

PR #1842, opened 2021-02-10
Initialized time=t before time marching in TransientReactiveTransport

mkaguer:
Fixing issue #1796


jgostick:
closing this in favor of a deeper refactor of transient algs, as discussed in meeting.  The preferred approach will remove the checking of steady-state completely, so algs will run until t_final no matter what.

PR #1844, opened 2021-02-11
Fix spacing issue for a network of two pores

### LINKED ISSUE #1732: network.spacing breaks for 1D networks with only two pores
```python
import openpnm as op
net = op.network.Cubic([2, 1, 1], spacing=1)
print(net.spacing)
```

Ni2M:
This PR closes #1732 

Sample code:
```python
import openpnm as op
import numpy as np
net = op.network.Cubic([1, 3, 1])
print(net.spacing)
```
output: 
```python
[0. 1. 0.]
# net = op.network.Cubic([3, 3, 3]) # still works for 3D networks
# [1. 1. 1.]
# net = op.network.Cubic([3, 3, 1]) # still works for 2D networks
# [1. 1. 0.]
```

### DIFF
--- a/openpnm/network/Cubic.py
+++ b/openpnm/network/Cubic.py
@@ -254,10 +254,14 @@ class Cubic(GenericNetwork):
         for ax in [0, 1, 2]:
             if dims[ax]:
                 inds = np.where(unit_vec[:, ax] == unit_vec[:, ax].max())[0]
-                temp = np.unique(mag[inds])
-                if not np.allclose(temp, temp[0]):
-                    raise Exception("A unique value of spacing could not be found")
-                spacing[ax] = temp[0]
+                if np.ndim(mag) != 0:
+                    temp = np.unique(mag[inds])
+                    if not np.allclose(temp, temp[0]):
+                        raise Exception("A unique value of spacing could not be found")
+                    spacing[ax] = temp[0]
+                else:
+                    temp = mag
+                    spacing[ax] = temp
         self.settings['spacing'] = spacing
         return np.array(spacing)
 
--- a/openpnm/network/Cubic.py
+++ b/openpnm/network/Cubic.py
@@ -139,9 +139,7 @@ class Cubic(GenericNetwork):
         elif connectivity == 6 + 8 + 12:
             joints = face_joints + corner_joints + edge_joints
         else:
-            raise Exception(
-                "Invalid connectivity receieved. Must be 6, 14, 18, 20 or 26"
-            )
+            raise Exception("Invalid connectivity. Must be 6, 14, 18, 20 or 26.")
 
         tails, heads = np.array([], dtype=int), np.array([], dtype=int)
         for T, H in joints:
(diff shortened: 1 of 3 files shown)


ma-sadeghi (changes requested):
Great, could you also add a unit test to catch the corner cases? (the line with `raise Exception`, and the lines after `else` clause.

Also, could you try it with networks with only two pores? In your example, you have 3 pores. Try shape = [1, 2, 1] or [2, 1, 1]

Ni2M:
Thank you Amin for review and comments. For sure. I will implement them.

Ni2M:
After reviewing the code, I think we have two raise exception that are related. To raise an error in the first exception the throat must not be aligned with the lattice (directional throat) which can happen by changing the pore coords "test_spacing_on_joggled_network". This exception also occurs when we add boundary pores (which results in len(c.keys()) >sum(dims)) unit test of "test_spacing_on_network_with_boundary_pores". Both of these cases fall in the category of raise error later in the code ("A unique value of spacing could not be found"), but they are already captured ealier in with the first exception. If we try following code, they will be captured by second exception but only because the first exception is commented.
``` python
import openpnm as op
import numpy as np
from openpnm import topotools
net = op.network.Cubic([2, 2, 1])
#net.add_boundary_pores()
print(net['pore.coords'])
P12 = net["throat.conns"]
net['pore.coords'][2]=[1.2,0.5,0.5]# instead of [1.5,0.5,0.5]
C12 = net["pore.coords"][P12]
mag = np.linalg.norm(np.diff(C12, axis=1), axis=2)
unit_vec = np.around(np.squeeze(np.diff(C12, axis=1)) / mag, decimals=14)
spacing = [0, 0, 0]
dims = topotools.dimensionality(net)
# Ensure vectors point in n-dims unique directions
# c = {tuple(row): 1 for row in unit_vec}
# if len(c.keys()) >sum(dims):
#     raise Exception(
#         "Spacing is undefined when throats point in "
#         + "more directions than network has dimensions"
#     )
mag = np.float64(mag.squeeze())
for ax in [0, 1, 2]:
    if dims[ax]:
        inds = np.where(unit_vec[:, ax] == unit_vec[:, ax].max())[0]
        if np.ndim(mag) != 0:
            temp = np.unique(mag[inds])
            if not np.allclose(temp, temp[0]):
                raise Exception("A unique value of spacing could not be found")
            spacing[ax] = temp[0]
        else:
            temp = mag
            spacing[ax] = temp
print(spacing)
```

In other words, I could not find any other case that does not raise error in the first exception but still raises error in the second exception. Because the condition of unit_vec will be checked in the first exception already. I would appreciate any comment related to this problem. Thank you.

Ni2M:
Current results to check 2 pores matches:
```python
import openpnm as op
import numpy as np
#net = op.network.Cubic([1, 2, 1])
net = op.network.Cubic([2, 1, 1])
print(net.spacing)

#[0. 1. 0.]
#[1. 0. 0.]
```

ma-sadeghi (approved):

PR #1846, opened 2021-02-15
Remove ReactiveTransport._set_variable_props

### LINKED ISSUE #1830: Remove `_set_variable_props`
This method doesn't do anything, and we're going to repurpose the name in V3

ma-sadeghi:
Closes #1830.

There are some scenarios where it might be used, like when doing shape optimization via an iterative process:

1. do some simulation
2. calculate some metric
3. modify psd/tsd
4. repeat until metric is optimized

This is basically what Tony's students are currently working on. So, it might not be entirely useless. Not sure if we should merge this PR.


jgostick:
I was actually thinking that the iterative props should be added a Mixin, since it's not always desirable to have this 'feature'

PR #1859, opened 2021-02-20
Allowing passing an axis handle to the plot_tools functions

jgostick:


ma-sadeghi:
How about making this PR a v3, and deprecate accepting the figure object altogether in all plotting functions. I think the convention is only accepting the axis.

PR #1893, opened 2021-03-30
Mayavi wrapper plot3d

hfathian:


ma-sadeghi:
Hey @jgostick @hfathian what does this PR do?

ma-sadeghi:
We should probably compile a short script out of this PR and put it in the currently non-existent contrib folder.

ma-sadeghi:
Copied to the private repo

PR #1899, opened 2021-04-13
Correct num_range of pore.seed model in 2D to be same as 3D class

### LINKED ISSUE #1898: Fix pore seed distribution in StickAndBall2D class
https://github.com/PMEAL/OpenPNM/blob/c01c5d3433c411bc69968feb7847cd2e3ba2eaa3/openpnm/geometry/StickAndBall2D.py#L178

Fixing StickAndBall2D class to have same pore seed as StickAndBall. View link above to see how num_range=[0.5, 0.5] but should be num_range=[0.2, 0.7] to be like StickAndBall. I am making PR to fix it.

mkaguer:
Fixes #1898


ma-sadeghi:
@mkaguer Two things:
1. Could you please look into the failed examples and see why they failed? If it's because of the change in seed numbers, then just rerun the whole notebook, save it and commit the changes. Since you only changed the seed numbers, it's probably that.
2. When linking a PR to an issue, you need to follow GitHub's strict guidelines, so merging the PR closes the issue(s). See [here](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). For instance, it should be fixes #issue_number rather rather than fixes issue #isse_number. Not that it's too important, but it's convenient if they're properly linked.

mkaguer:
Hi @ma-sadeghi I updated and saved changes to the notebook but the error still appears. Is the testing check expecting the outputs from the workbook to be the same before the pore.seed change. Thanks.

ma-sadeghi:
> Hi @ma-sadeghi I updated and saved changes to the notebook but the error still appears. Is the testing check expecting the outputs from the workbook to be the same before the pore.seed change. Thanks.

Thanks. I think you need to set the random seed number at the top of the notebook via `np.random.seed(10)`, then rerun the entire notebook locally, save it, and then push the changes.

ma-sadeghi:
@mkaguer Also, ignore the Windows build for now. There seems to be an issue with GitHub's Windows instances and MKL (#1907)

mkaguer:
> > Hi @ma-sadeghi I updated and saved changes to the notebook but the error still appears. Is the testing check expecting the outputs from the workbook to be the same before the pore.seed change. Thanks.
> 
> Thanks. I think you need to set the random seed number at the top of the notebook via `np.random.seed(10)`, then rerun the entire notebook locally, save it, and then push the changes.

Ok yes, makes sense.

PR #1902, opened 2021-04-16
Stop exporting peclet numbers

### LINKED ISSUE #1901: Remove lines that store peclet numbers
https://github.com/PMEAL/OpenPNM/blob/c01c5d3433c411bc69968feb7847cd2e3ba2eaa3/openpnm/models/physics/ad_dif_mig_conductance.py#L175

https://github.com/PMEAL/OpenPNM/blob/c01c5d3433c411bc69968feb7847cd2e3ba2eaa3/openpnm/models/physics/ad_dif_conductance.py#L97

These lines of code can be removed. No need to export peclet numbers. Also, was creating an issue when you had more than one physics object adding ad_dif_conductance since peclet numbers were getting overwritten.

mkaguer:
Fixes #1901

### DIFF
--- a/openpnm/models/physics/ad_dif_conductance.py
+++ b/openpnm/models/physics/ad_dif_conductance.py
@@ -94,10 +94,6 @@ def ad_dif(target,
     Peij[(Peij < 1e-10) & (Peij >= 0)] = 1e-10
     Peij[(Peij > -1e-10) & (Peij <= 0)] = -1e-10
 
-    # Export Peclet values (half only since Peij = -Peji)
-    phase['throat.peclet.ad'] = _np.nan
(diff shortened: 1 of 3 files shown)


mkaguer:
@ma-sadeghi in \OpenPNM\tests\unit\core\ModelsTest.py line 5 there is this `from testfixtures import LogCapture` which is unused. Could we delete that? Also, it can't actually find the module testfixtures. Is it like that on your computer as well?

ma-sadeghi:
> @ma-sadeghi in \OpenPNM\tests\unit\core\ModelsTest.py line 5 there is this `from testfixtures import LogCapture` which is unused. Could we delete that? Also, it can't actually find the module testfixtures. Is it like that on your computer as well?

I think it's safe to delete. This PR looks good. I'm merging it. Thanks :)

PR #1914, opened 2021-05-18
Replaced sp.sparse with sprs in graphtools

mkaguer:
I read [here](https://stackoverflow.com/questions/8696322/why-does-this-attributeerror-in-python-occur) that submodules (such as scipy.sparse) don't automatically get imported and so we need to use `import scipy.sparse as sprs`. We already had this command at the top of script. Simply replace sp.sparse with sprs.

### DIFF
--- a/openpnm/core/Base.py
+++ b/openpnm/core/Base.py
@@ -247,7 +247,7 @@ class Base(dict):
             if self._count(element) == 0:
                 self.update({key: value})
             else:
-                raise Exception('Provided array is wrong legth for ' + key)
+                raise Exception('Provided array is wrong length for ' + key)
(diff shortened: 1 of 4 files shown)


mkaguer:
Line 297 gave me the original error: AttributeError: 'module' object has no attribute 'sparse'

But went back and replaced all instances of sp.sparse with sprs and removed `import scipy as sp` because it was no longer used.

mkaguer:
@ma-sadeghi some checks are being cancelled. I am not sure why...

ma-sadeghi (approved):
Thanks for the PR @mkaguer, looks good!

PR #1947, opened 2021-07-15
Remove 2D methods in Poisson, Electroneutrality, and Laplace

### LINKED ISSUE #1753: 2d and 3d versions of electroneutrality, Poisson, and Laplace models are identical
I think we should get rid of the 2d versions.

Ni2M:
This PR closes #1753 by removing 2D version of methods. The only difference bw 2D and 3D was using pore/throat diameter as area in ionic conductance and pore/throat area as volume in source terms.


Ni2M:
Discussion for dev meeting:

We removed 2D methods and included 2D models within their general 3D method by checking the network dimension. 
However, there's a concern that the user may want to use a 3D model for a 2D network! We could just remove the 2D assignments based on the network dimension and alternatively pass pore.diameter and throat.diameter for areas in ionic conductance which gives the user control. 

However, there's one place in charge conservation (https://github.com/PMEAL/OpenPNM/blob/bdb1daac683aa1192ad5f826c0d644395e0adde6/openpnm/models/physics/source_terms.py#L77)  that 2D and 3D are distinguished and our automatic assignment based on network dimension is more noticeable and helpful.

jgostick:
The changes in this PR are not actually needed for V3, which will have the 2D/3D info embedded in the geometry objects, so the downstream code (i..e. algorithms) don't have to think about it.

PR #1954, opened 2021-07-23
Simplify custom class instantiation

jgostick:
This PR changes how models are defined for the subclasses, by putting them in a dictionary instead of using ``add_model`` within the class definition.


jgostick:
I'm redoing this in new PR

PR #1974, opened 2021-08-30
Removing setup method from transport algs

jgostick:


Ni2M:
Problems with metrics relperm algorithm are fixed now.

PR #1977, opened 2021-09-06
Create a maze solver

TomTranter:
Hey all, I saw this post that Amin commented on Linkedin and felt inspired to give it a go. Funny thing is I had to calculate rate in the throats and then patch the data to the pores to get it looking right... Not sure why. If we could approve this and post a comment back to Linkedin that would be cool


TomTranter:
By the way @jgostick this got me thinking about your dead-ended pore rants of yesteryear. Might be a nice way to demonstrate it. I generated the mazes here https://www.mazegenerator.net/

ma-sadeghi:
Hey Tom, this is great! I really enjoyed it. 

I added some text to your notebook, and also added an alternative approach (using network extraction) at the end of the notebook. Take a [look](https://github.com/PMEAL/OpenPNM/blob/maze_example/examples/notebooks/fun/maze_solver.ipynb).

For sure go ahead and post it back, but I think maybe it's worth it if you also do a separate post in addition to a comment on Jeremy Theler's post.

ma-sadeghi (approved):

TomTranter:
hey @ma-sadeghi that's great I like what you added and the network shows a nice speed up. Do you think we should add the lines to install openpnm and porespy in the notebook as most people won't have these installed by default.

TomTranter:
I guess the graph purists are bound to argue that we could have also used Dijkstra's algorithm to find shortest path so maybe we should add some comments that yes we know this is a fun algorithm and you can also do lots of other cool physics on graphs with all our algorithms

jgostick:
I think when we do V3, we'll make a 'contrib' folder in the examples somewhere, and this should go there, but now it's great.  I'll merge it.

Issue #1912, opened 2021-05-05
new shape factor models break on networks with subdomains


jgostick:
It seems the indexing did not accounting for the possibility that a geometry might not apply to all pores in the network.

Ni2M:
This issue was addressed in PR #2001.

Ni2M:
This issue was addressed in a test in PR #2001 where only some throats were assigned to a geom. I will double check with another test with more subdomains in PR #2032.

Ni2M:
I double checked with another multidomain test in PR #2032 (test_generic_diffusive_multi_domain) that could also pass in the merged PR #2001. Therefore, the current version of dev works fine on subdomains. This issue is now closed.

PR #2001, opened 2021-09-10
Making ionic/elec/Thermal conductance consistent with the new size factor model

### LINKED ISSUE #1753: 2d and 3d versions of electroneutrality, Poisson, and Laplace models are identical

Ni2M:
This PR closes  #1753. 
To do list:
- [x] Clean up ionic conductance models
- There are many unnecessary inputs, etc
- Laplace and Poisson are the same models
- [x] Remove 2D models (2D networks already have size factors, no need for a separate phys model!)
- [x] Incorporate size factor in the following models:
 - ionic conductance
 - Thermal conductance
 - Electrical conductance
 (suggestion/discussion for dev meeting: maybe we need a generic series_resistors model that takes in the props (whether electrical/diffusive/thermal conductivity) for simple phys models instead of defining a series_resistors for each case.)
- [x] Checks, Tests and docstrings
- [x]  Double check the electroneutrality_generic
- [x] There's no test for electroneutrality. Maybe we need to add one? -Test added


jgostick:
- [x] Remove try/excepts for property lookups on phase objects since interpolation happens automatically[^1]
 [^1]: check if it's only the case for pores to throats or throats to pores. + lines without try,except:
checked: either pore or throat, they both are checked for interpolation in GenericPhase once the phase is instantiated.
- [x] Ensure that [throats] indices are used consistently throughout, so either at the very end (i.e. return vals[throats]) or in all lines above.  
- [x] Update files form this PR, but also for the other conductance models, which would break on subdomains.  I think this is related to issue #1912.  [^2] 
[^2]: Updating other conductance models: ordinary diff, mixed diff, etc to accept size factor (instead of shape factor) will be a separate PR (don't do those changes here) That's issue #2009.
- [x] Create a utils.py file in the models/physics folder and put ``generic_transport_conductance`` there, then call it from each of the "wrapper" models (i.e. ``generic_diffusive``) [^3]
[^3]: We can't use the generic_hydraulic as a wrapper, because viscosity is in the denominator.
- [x] Rename all these new generic models to: ``generic_<physics>`` thus ``generic_diffusive``
- [x] Update advection_diffusion and advection_diffusion_migration as well [^4]
[^4]: These models get the conductances as the input, therefore they don't need changes based on size factor. The reason is before calling ad_diff and ad_diff_mig, the geometry diffusive factor model and the physic diffusive and hydraulic conductances must be called. So the new generic_<phys> method has already been taken care of in those phys models.
- [x] Delete dispersive_conductance.py file
- [x] add required tests

Ni2M:
I think this PR is ready to merge. The issue #2009 that follows the changes applied here will be addressed in a separate PR.

PR #2004, opened 2021-09-14
Adding a new plot_network to topotools

### LINKED ISSUE #1887: Quick visualization with matplotlib is annoyingly limited
Plotly offers 3D visualization of graphs:  https://plotly.com/python/v3/3d-network-graph/

Ni2M:
This PR closes #1887. The method plot_network_interactive uses plotly to plot a 3D network. More details are included in the docstring.

### DIFF
--- a/openpnm/topotools/__init__.py
+++ b/openpnm/topotools/__init__.py
@@ -63,5 +63,4 @@ from .plottools import plot_tutorial
 from .plottools import plot_connections
 from .plottools import plot_coordinates
 from .plottools import plot_networkx
-from .plottools import plot_vpython
-from .plottools import plot_network_interactive
+from .plottools import plot_network_jupyter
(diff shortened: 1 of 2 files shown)


Ni2M:
```python
size=10
net = op.network.Cubic(shape=[size, size, size])
geo = op.geometry.StickAndBall(network=net, pores=net.Ps, throats=net.Ts)
op.topotools.plot_network_interactive(net)
```
Snapshot of the result:
image

ma-sadeghi:
@Ni2M Looks good. Is there a way to size the throats? (I guess not)

Ni2M:
Thanks. That's a good point. Not that I know of. For throats there's the thickness of the line that is only a scalar. I tried using throat diameter instead,  but it gave an error that it must be a scalar, not an array.

jgostick:
yes, the throats cannot be sized separately, only as a whole group.  I did have some trouble with their coloring though. The z-directional throats showed a gradient in color while the horizontal throats are all basically the same color.  I think Niloo's pic is showing this too.  Should we investigate now, or merge and fix later?

Ni2M:
That looks nice. I double checked the throats' coloring. My results also showed only z dir throats' different colors and labels.

PR #2026, opened 2021-09-29
Clean up network classes

jgostick:


jgostick:
This was too ambitious all at once.  Moving useful code to a new PR.

PR #2030, opened 2021-10-04
Updating math functions in misc to accept an array of props instead of individual args

### LINKED ISSUE #1919: product, sum, and othe math models should accept a list of props, not just 2 props
like

```
obj.add_model(propname='pore.blah', model=op.models.misc.product, props=['pore.A', 'pore.B', 'pore.C', etc]
```

Ni2M:
This PR closes  #1919. Currently, the product function accepts the arguments separately. Current product function can multiply more than two arguemnts (accepts **kwarg), but they all need to be passed individually. The function should accept the props in a list.
To do:
- [x] change the function
- [x] change in the examples,etc
- [x] test file (current test file already tests the product function)
Fixed this problem: Notes: the dependency list will be changed. Probably related to finding the model prop in dependency grpah (now we have props as an array). Looking into the problem.
Problem:
```python
import openpnm as op
import numpy as np
net = op.network.Cubic(shape=[3, 3, 3])
geom = op.geometry.GenericGeometry(network=net,
                                   pores=net.Ps)
geom['pore.max_size'] = 0.4

geom.add_model(propname='pore.seed',
               model=mods.misc.random,
               element='pore',
               num_range=[0, 0.1],
               seed=None)

geom.add_model(propname='pore.diameter',
               model=mods.misc.product,
               props=['pore.max_size', 'pore.seed'],
               regen_mode='normal')

geom.add_model(propname='pore.volume',
               model=mods.geometry.pore_volume.sphere,
               pore_diameter='pore.diameter',
               regen_mode='normal')


geom.add_model(propname='pore.area',
               model=mods.geometry.pore_cross_sectional_area.sphere,
               pore_diameter='pore.diameter',
               regen_mode='normal')


tree = np.asarray(geom.models.dependency_list())
print(tree)
# output: ['pore.diameter' 'pore.area' 'pore.seed' 'pore.volume']

# desired output: ['pore.seed' 'pore.diameter' 'pore.area' 'pore.volume']
```

### DIFF
--- a/openpnm/core/ModelsMixin.py
+++ b/openpnm/core/ModelsMixin.py
@@ -114,8 +114,13 @@ class ModelsDict(PrintableDict):
             # Filter pore/throat props only
             dependencies = set()
             for param in self[model].values():
-                if is_valid_propname(param):
-                    dependencies.add(param)
+                if type(param) == list:
+                    for element in param:
+                        if is_valid_propname(element):
+                            dependencies.add(element)
+                else:
+                    if is_valid_propname(param):
+                        dependencies.add(param)
             # Add depenency from model's parameters
             for d in dependencies:
                 if not deep:
--- a/openpnm/geometry/CirclesAndRectangles.py
+++ b/openpnm/geometry/CirclesAndRectangles.py
@@ -78,8 +78,7 @@ class CirclesAndRectangles(GenericGeometry):
 
         self.add_model(propname='pore.diameter',
                        model=mods.misc.product,
-                       prop1='pore.max_size',
-                       prop2='pore.seed')
+                       props=['pore.max_size', 'pore.seed'])
 
         self.add_model(propname='pore.volume',
                        model=mods.geometry.pore_volume.circle,
--- a/openpnm/geometry/ConesAndCylinders.py
+++ b/openpnm/geometry/ConesAndCylinders.py
@@ -82,8 +82,7 @@ class ConesAndCylinders(GenericGeometry):
 
         self.add_model(propname='pore.diameter',
                        model=mods.misc.product,
-                       prop1='pore.max_size',
-                       prop2='pore.seed')
+                       props=['pore.max_size', 'pore.seed'])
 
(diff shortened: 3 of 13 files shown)


Ni2M:
The problem with the dependency change was related to https://github.com/PMEAL/OpenPNM/blob/268535fbcbbf90f73c13b8a90e54e6f6965ad5c6/openpnm/utils/misc.py#L544 where with the new approach it return False. The reason is for example now we have props=['pore.max_size', 'pore.seed'] which is a list (not a str). So when calling the dependency functions (graph/list/etc), props is ignored in is_valid_propname and not added to the correct order in the graph.

Codes that show the cause of the problem:
image

Ni2M:
Solution: 
Adding extra loop in the dependency_graph to check is_valid_propname for each element in a list. This fixes the problem for props that contain a list of valid props:
```python
import openpnm as op
import numpy as np
import openpnm.models as mods
net = op.network.Cubic(shape=[3, 3, 3])
geom = op.geometry.GenericGeometry(network=net,
                                   pores=net.Ps)
geom['pore.max_size'] = 0.4

geom.add_model(propname='pore.seed',
               model=mods.misc.random,
               element='pore',
               num_range=[0, 0.1],
               seed=None)

geom.add_model(propname='pore.diameter',
               model=mods.misc.product,
               props=['pore.max_size', 'pore.seed'],
               regen_mode='normal')

geom.add_model(propname='pore.volume',
               model=mods.geometry.pore_volume.sphere,
               pore_diameter='pore.diameter',
               regen_mode='normal')


geom.add_model(propname='pore.area',
               model=mods.geometry.pore_cross_sectional_area.sphere,
               pore_diameter='pore.diameter',
               regen_mode='normal')


tree = np.asarray(geom.models.dependency_list())
print(tree)
# output: ['pore.max_size' 'pore.seed' 'pore.diameter' 'pore.area' 'pore.volume']
```

Ni2M:
I think this PR is ready to merge.

PR #2031, opened 2021-10-04
Changing Voronoi class to accept only points argument

### LINKED ISSUE #1694: Voronoi classes should just accept points, not num_points
If points is an array, then behave as before, if points is a scalar, then generate points as is now down for num_points.

Ni2M:
This PR closes #1694. The Voronoi class will accept only points argument.
- [x] Change voronoi and sub_functions
- [ ] Test, docstrings


Ni2M:
This PR was addressed before. I am closing the PR.

PR #2032, opened 2021-10-05
Revisist conductance models to accept size factors instead of shape factors

### LINKED ISSUE #2009: Revisit all existing pore-scale conductance related models to work with new size_factor approach
Example: taylor_aris_diffusion requires knowing some size info (specifically area) but this is now embedded in the shape_factor models, so we need to figure out how to compute this value.  

Basically many of the more complicated pore-scale models probably still need size info that the geometry no longer provides.

Ni2M:
Fixes #2009


Ni2M:
New changes start here.
Notes: classic ordinary diffusion only needed throats' shape factor and not the pores' shape factors.
Note: tests of conductance are based on the idea that diffusive_size_factors must have already been defined (previously with shape factor we could use SF=1 in these cases, but now the geometrical info (integral of A/L) is included in size factor).

Ni2M:
Discussions for Dev meeting:
- [ ] Should we remove Classic Hagen Poiseuille method, as it is similar to using cylinders in series method in geom size factor and calling Hagen Poiseuille. However, there is a slight difference that Classic Hagen Poiseuille applies shape factor only to throats.

- [ ] With current changes the user "Must" specify the size factor then call each conductance method. Previously, size factors were included in the Shape factor so we had a default of SF=1 if the user doesn't define shape factor. However, size factors contain geometrical info (A/L or A^2/L , etc) and we can't define a default without adding additional input args to the conductance method (besides I think the goal of this PR was to reduce the number of input args). One solution may be to always call the cylinders in series size factors within the conductance methods, if no size factor was defined by the user. Another solution maybe to define a generic_size_factor model that only calculate A/L or A^2/L from pore info (pore area and length) instead of the integral so that it works as if the shape factor was 1. (SF=L/A*size factor)

- [ ] Valvante-Blunt hydraulic conductance uses a separate method and shape factor concept, so we didn't update it based on the size factors. Assumming the user will define the shape factors themselves SF=pore_area / perimeter **2.

- [ ] Conductance with zero_length_throat in this new approach is not applicable, as the size factor methods find the conduit length internally within their method using ```openpnm.models.geometry.conduit_lengths```. I would suggest updating size factors models to account for zero length, zero diameter, and such exceptions. These exceptions were handled in conductance models previously (e.g. [here](https://github.com/PMEAL/OpenPNM/blob/743687c42a23ad4ef072b7ca248cc0480b85cf0e/openpnm/models/physics/diffusive_conductance.py#L424)). But with unifying all conductance models and trying to remove geometrical input args in the conductance models, I would suggest handle these exceptions in the size factors.
The following shows the problem:
``` python
# DISCUSSION ON ZERO LENGTH THROAT TEST IN DIFF CONDUCTANCE
import openpnm as op
import openpnm.models.geometry.conduit_lengths as _conduit_lengths
net = op.network.Cubic(shape=[3, 3, 3])
geo = op.geometry.GenericGeometry(network=net,
                                  pores=net.Ps,
                                  throats=net.Ts)
geo['pore.diameter'] = 2.2
geo['throat.diameter'] = 0.56
phase = op.phases.GenericPhase(network=net)
phase['pore.permittivity'] = 78
phys = op.physics.GenericPhysics(network=net,
                                 phase=phase,
                                 geometry=geo)

mpo2 = op.models.geometry.diffusive_size_factors.spheres_and_cylinders
geo.add_model(propname="throat.diffusive_size_factors",
              model=mpo2)
# calcs for a conduit
print('throat size factor=',
      geo['throat.diffusive_size_factors.throat'][0])  
# instead of inf it is throat size factor= 246300864041439.78
g1 = 1.3*geo['throat.diffusive_size_factors.pore1'][0]
g2 = 1.3*geo['throat.diffusive_size_factors.pore2'][0]
gt = 1.3*geo['throat.diffusive_size_factors.throat'][0]
gtotal_large_gt = 1 / (1 / g1 + 1 / gt + 1 / g2)
gtotal_inf_gt = 1 / (1 / g1 + 1 / g2)
print('g inf=', gtotal_inf_gt, ',g large val=', gtotal_large_gt)
#g inf= 4.58028497910196 ,g large val= 4.5802849791018945
#  The reason is that geom.diff_size_factor internally uses
#  get_cond_length as follows for example:
L1, Lt, L2 = _conduit_lengths.spheres_and_cylinders(geo).T
geo['throat.conduit_lengths.pore1'] = L1
geo['throat.conduit_lengths.throat'] = Lt
geo['throat.conduit_lengths.pore2'] = L2
print('throat length=', Lt[0])  # instead of zero it is 1e-15
```
- [ ] Another problem related to the previous note on zero_length_throat: currently boundary pores and conduits models fail and conductances are NaN because these exceptions are no longer handled in the conductance models. So I would suggest adding these exceptions to the size factors where we already have the geometrical info.

- [ ] With the new approach Standard Physics doesn't have the shape factor models. The default shape factor method in Standard physics was conical_frustrum_and_stick equivalent to "cones_and_cylinders" in geometry size factors. Therefore, we added a cones_and_cylinders to the stick and ball geometry. It is not the best option to add a "cones" model to stick and ball, but due to possible problems with the "sphere_and_cyliner" dimensions (D/L ratio) limitation, we used cones_and_cylinders instead. I think we need to find a better solution.

Ni2M:
I think this branch is ready to merge! Please not following:
1) There is no shape factor and no associated function/assigning shape factor (except for valvante blunt method), as things are now accepting size factor.
2) We had to keep cones_and_cylinders size factor in stick and ball to pass the tests (as their default Standard phys shape factor was conical frustrum and stick before). In a separate branch the Standard phys and its associated tests are fixed.
3) We had to remove boundary geom here to pass the tests. This geom is removed in a separate branch as well.
4) In a separate branch Stick and ball, conduit length, endpoints are removed  and used for test files (_stickAndball)

ma-sadeghi:
I started making similar changes to some other branch, and now it's all over the place. Sorry for that. Let me consolidate both changes before merging this PR as I've made some other changes to make the tests more robust (ex. not invoking pore-scale models inside unit tests, but rather hard coding dependencies)

Ni2M:
Hi @ma-sadeghi. I appreciate all your time and considerations. I think we had to apply similar changes in different branches as  they were inter-related. It would be great to consolidate them and as you mentioned the order of the branches to be merged matter now and we may need to apply some final changes in each PR as we are merging them one by one. Please let me know if I should apply any changes when it's needed. I can also adjust my changes here based on your branch once that is merged. Thank you.

ma-sadeghi:
Thanks @Ni2M, this looks great!

Ni2M:
Thank you @ma-sadeghi  for your help and great job on fixing the issues. It's great!

xu-kai-xu:
@Ni2M, hi, I am confused by the `hydraulic_size_factors` and `shape factors`. As you mentioned, I remember in v2.5, there is no need to calculate `hydraulic_size_factors`, e.g., the classicall H-P equation. In that situation, the `shape factor` is 1 for all pores and throats, am I right? I want to know in this case, what is the value of hydraulic_size_factor. emmm, I can't distinguish the two fators' differences. Thanks for your help.

Ni2M:
@xu-kai-xu Hi. The size_factors are the geometrical part of the conductance and with this new change the conductance models take in the size_factor and multiply that by the inverse of viscosity. Note that the the effect of shapes are already included in the size_factors so the final result of the conductance models would be the same as using shape factors. We've moved the calculations related to the effect of shapes in the geometry models (size factor models) and removed the geometry/shape-related calculations in the conductance models (they are now included in size factor models).
Here are notable differences:
1) Size factors models include more shapes whereas shape factor models included only spheres/cylinders/conical frustums.
2) Size factors don't have a default of 1 unlike shape factors because size factors themselves include the geometrical/shape-related part of the conductance calculations (which were previously included in the shape factor and conductance models).

Example: 
Assuming Hagen-poiseuille for example and stick and ball shape assumption: Q=**(A^2/L*8*Pi)***SF*(1/Mu)*DeltaP.
With previous approach we had these steps:
1) Calculate Shape_factor SF=(L/A^2)*(1/F)    #this F includes information on the deviation of shape and geometry (F is INTEGRAL(1/A^2) dx)
So the default shape_factor=1 meant that our element is a cylinder with constant circular cross section.
2) Calculate conductance=(A^2/L*8*Pi)*(SF )*(1/Mu) and SF=1 again meant the element is a cylinder.
If the SF was calculated, we'd have: conductance=(A^2/L*8*Pi)*(SF )*(1/Mu)=(A^2/L*8*Pi)*(L/A^2)*(1/F)*(1/Mu)=(1/8*Pi)*(1/F)*(1/Mu)=(geometrical and shape info)*(1/Mu)

Now with new changes the idea of the size factor is to calculate size_factor=(geometrical and shape info) depending on geometrical assumptions for pores/throats shape (cylinders/spheres/cubes/etc) in geometry models. This means the user must add a size factor model before calling the hydraulic conductance methods. 

Now with the changes for the same example and spheres and cylinder assumption we have these steps:
1) Calculate size factor S=(1/8*pi)*(1/F) #this F includes information on the deviation of shape and geometry (F is INTEGRAL(1/A^2) dx)

2) Calculate conductance=S *(1/Mu)

Note that for each shapes the calculations of S (size factor) are different. As an example I mentioned the steps for spheres and cylinders. We will be updating our examples and adding the explanations in a notebook in our next release.

PR #2036, opened 2021-10-06
Developing new Base class(es)

jgostick:
This is a VERY significant change to the backend of OpenPNM.  It completely re-thinks how data is managed, and solves quite a lot of issues/headaches/painpoints we have.  It is almost identical to the user, functionality-wise, but there are a few breaking changes.


jgostick:
This has turned into a disaster.  I was many hundreds of lines in, with no end in sight, and playing whack-a-mole with each new line.

PR #2073, opened 2021-10-20
Remove StickAndBall

### LINKED ISSUE #2005: Stop using StickAndBall, except in tests (for now)
We are removing the StickAndBall class from the geometry module and replacing it with SpheresAndCylinders.  This means we need to do a 'find and replace' in all the unit tests. There will be a LOT of broken tests that need fixing after this.  To prevent this from happening again in the future we should NOT use a geometry class in the tests, but instead just use random numbers for things like conductance and pore size, setting the number generator seed to a constant starting point of course.

### LINKED ISSUE #2069: Remove conduit_lengths and endpoint models
These are no longer needed and are confusing.

Move them to the StickAndBall file...then inside the class you'd do:

``` python
class StickAndBall(...):
    def __init__(self, ...):
        self.add_model('throat.endpoints', model=endpoints)

# --- end of class ---

def endpoints(...)
    <code>

```

jgostick:
Fixes #2069 
Fixes #2005


Ni2M:
I think this branch is ready to merge!
Please note following:
1) We didn't replace stickandball with spheresandsylinders model in [example.py](https://github.com/PMEAL/OpenPNM/blob/c45a26e9ffda0447ac2bc8f19ccda399d0a4ed5a/example.py#L12) and some .py files that aren't test files, as spheresandsylinders does not include information on pore/throat area and conduit length which are used in the current version of Standard physics. Once the Standard phys new size factor PR is merged, we can just replace stickandball in those example/files with sphersandcylinders.
2) Throat endpoints are removed from geom models. However plottools, STL, salome.py files use endpoints.

PR #2087, opened 2021-10-22
Allowing obj['param.value'] = 1.0

### LINKED ISSUE #2035: Allow 'param' prefix in the object dictionaries
The new Base refactor will allow users to write arbitrary data to their objects using the 'param' prefix.  

Note that we considered allowing arbitrary prefixes, but realized this won't work since OpenPNM needs to know what each means.  For instance, if the user creates 'bond.data', how can we deal with it during trimming of pores and throats?  OpenPNM would not know that it's a 'throat' property.  Of course there are ways to defines alias, etc, but this adds headache we don't need.  So we're sticking with 'pore', throat' and now 'param'.  

That being said, we will attempt to make the code flexible so that new prefixes "could" be added in future, so will try not to hard code the 'pore/throat' prefix.

jgostick:
This is an interesting approach to adding params...
I created a mixin (my new favorite!) that adds a params attr to the object, and also overloads thet set/get item method to intercept any keys that start with param, then directs them to read/write to the params dictionary. 

Not sure if this is the way to go (mixin overload?) but it feels right.

Fixes #2035


jgostick:
* Enhance __getitem__ to look up values on domain if not found on self
* Add ``params`` as a method which works similar to ``props`` and ``labels``, and move params dict to ``_params``
  * If we do this, we lose the ability to ``print(pn.params)``.
  * I guess we should add it to the ``__str__`` method so ``print(pn)`` includes an extra level for params
* Remove checks for 'param's in setitem and elsewhere that lingered

PR #2092, opened 2021-10-22
More refactoring of base

### LINKED ISSUE #2042: Remove map_pores/throats and replace with to_global and to_local
The mapping functionality is only used very occassionally, but needs to be kept.  I think the terminology is confusing though and propose we add ``to_global`` and ``to_local``.  These would complement the ``to_mask`` and ``to_indices`` functions that we have (also note their names have an underscore added for better readability).  

Additionally, the ``pores`` and ``throats`` methods will have a ``relative=True/False`` argument added.  Although, as I write this I'm seeing a mis-match between the terminology.  I guess the argument should be ``local=True/False`` for consistency.  OR, the methods could be called ``to_relative`` and ``to_absolute``?

jgostick:
Adding ``to_global`` and ``to_local`` to Base to replace map_pores/throats

Fixes #2042

### DIFF
--- a/openpnm/core/Base.py
+++ b/openpnm/core/Base.py
@@ -639,6 +639,29 @@ class Base(dict):
         indices = self._parse_indices(mask)
         return indices
 
+    def to_global(self, pores=None, throats=None):
+        if pores is not None:
+            element = 'pore'
+            locs = pores
+        elif throats is not None:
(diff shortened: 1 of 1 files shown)


jgostick:
I need this functionality on another PR, so merging

PR #2097, opened 2021-10-23
Fix add/drop locations

### LINKED ISSUE #1858: add/drop locations not quite working as advertized
The docstring says that if pores/throats are already on a geometry then they will be changed, and they will be removed from the physics.  This does not seem to be happening. I was able to remove them from the geom add them to a new geom, but not 'switch' them directly.  I'm thinking that we should revert to ``set_locations`` as a public method, with modes in V3.0.

### LINKED ISSUE #2096: Add domain and subdomains attrs
If we had a ``subdomains`` attr on network and phase objects, and a ``domain`` attr on geometry and physics, life would be a LOT easier.

For instance, in the _set_locations function on Subdomain objects, if we could ask the ``boss`` for it's list of ``subdomains`` is would be much clearer and much less code.  

I'm going to add these as hidden methods for now, and we can expose later if necessary.

### LINKED ISSUE #2101: Add physics attr to GenericGeometry and geometry attr to GenericPhysics
These are just shortcuts for the equivalent methods on the project but are much more readable and quite handy.

For instance:
``geoms = phys.project.find_geometry(physics=self)``
and 
``physics = geom.project.find_physics(geometry=self)``

becomes:
``geoms = phys.geometry``
and
``phys = geom.physics``

The entire thing is done using the 'top-down' look-up so has no object attachment and the potential for memory leaks.

jgostick:
Fixes #1858 
Fixes #2096 
Fixes #2101

This PR is not just a refactor but has some api changes as well.  ``add/drop_locations`` is removed, and ``set_locations`` is now public.  It will be possible to set the physics pores/throats using either a geometry arg or pores & throats.  Also adding many tests for subdomain consistency which seems to have gone missing.


jgostick:
This PR adds quite a bit or new 'attrs' to the various classes.  We should discuss if this is actually a good idea before merging. There is also a decision required RE how much freedom we want to give the user to set locations manually.

PR #2102, opened 2021-10-24
Fix nernst planck examples

### LINKED ISSUE #2088: Nernst-Planck example scripts are broken (both 2d and 3d)
How did this happen? Don't we test the `scripts/` folder?

mkaguer:
Fixes #2088


ma-sadeghi:
@mkaguer Hmm, so basically, reducing the number of iterations from 100 to 50 fixes the issue. This seems a bit strange to me. Why is it that if we continue iterating, instead of getting closer to the solution, we're moving away?

mkaguer:
Hi @ma-sadeghi. As discussed the issue is using 'electroneutrality' assumption. I am pushing changes so that 'poisson' is used in this example. The user should be cautioned to use electroneutrality.

ma-sadeghi:
Hey @mkaguer, the 3d version of the example works fine, but the 2d one still diverges. Could you please look into it? Thanks!

mkaguer:
Hi @ma-sadeghi. I think I found the problem. The StickAndBall2D class was adding sphere and cylinder size factors. Shouldn't we be using circles and rectangles for the 2D case? I changed that and now it converges okay. Please let me know if this is not right.

Issue #2077, opened 2021-10-21
__setitem__ on SettingsDict should look for certain keys and take specific actions


jgostick:
For instance, when setting 'conductance' on an algorithm, we should also set _pure_A to None.  This forces the algorithm to rebuild A using the new conductance.  As it is now, it just reruns the code using the cached A.  

The bigger picture here is that maybe we need to subclass the SettingsDict for specific purposes.  For instance, catching 'conductance' is only relevant for algorithms, so it's __setitem__ method should be overloaded accordingly, but the settings on a network do not need this.

jgostick:
In our new overhaul of settings, we use the ``traits`` package, which lets us do some very useful things.  At the moment it only enforces the type of each setting, so that you can't write an object when a string is expected.  However, it offers additional functionality: the ability to observe changes in a setting and trigger some response.  I have not explored this yet, but once the PR is finalized and merged I can look into it.

jgostick:
This should have been closed with PR #2121

PR #2121, opened 2021-11-01
Convert setting dict to class w attrs

### LINKED ISSUE #1951: alg.settings['sources'] should be a set, not a list
We want `alg.settings['sources']` to be unique, so why not change its data type to `set`? This gets rid of some manual checks we do here and there, for instance:

https://github.com/PMEAL/OpenPNM/blob/bdb1daac683aa1192ad5f826c0d644395e0adde6/openpnm/algorithms/ReactiveTransport.py#L219-L220

Clearly, not urgent at all, just a note for future.

### LINKED ISSUE #2108: Rethink the settings attribute
Does it still need to be a dictionary?  Why not just a dataclass (or basic object) with attrs, like:

``` python
alg.settings.tol = 45
alg.settings.max_iter = 10
print(alg.settings)
<output some nicely formatted display>
```

If don't this way, we could also add setter/getter property methods to protect some settings (i.e. to ensure that max_iter < 100)

I **think** the reason we had it as a dict before was to do ``obj.settings[variable_handle] = value`` in the ``setup`` method, instead of ``setattr(obj.settings, variable_handle, value)``...the dict lookup is much cleaner for sure. However, we have gotten rid of the ``setup`` method so there is literally nowhere in the code where we're using an object handle as a dict key.

jgostick:
This PR is an attempt to tighten-up the ``settings`` attribute on classes.  Some of the details of the current settings machinery are outlined in #2108.

Fixes #1951 
Fixes #2108 

Also related to #2077

---

Here is a list of what we want:

* Should be initialized like dataclasses, with settings listed as class attributes.  This is the easiest and clearest way to specify such things.  It would also be possible to specify these as a dict or in a json file, but that would hinder the next requirement:

``` python
class Settings:
    a = 1
    b = ['foo', 'bar']
```

* Needs to be a class so that docsstrings can be included as below.  This seems to be the only place where a meaningful explanation of the setting can be included, and made visible to the user via the IDEs help panel (ie. jupyter or spyder). Even if there were some other place to put these docstrings, having them in the class defition enables the next requirement: 

``` python
class Settings:
    r"""
    a : int
        A setting that controls the weather
    b : list of strings
        The names to call things
    """
    a = 1
    b = ['foo', 'bar']
```

* Inheriting and augmenting docstrings from parent classes is essential since each level of subclass also adds a new level of settings.  We use docrep to inherit and collate the docstrings between objs, so this sort of class-based docstrings works perfectly with it.

* We would like to implement some protections of the attributes RE datatypes.  Python dataclasses let you protect the type upon intialization, but afterwards you can write any old things to any attribute.  There are two packages that let you do this: attrs and traits.  The latter is very slick:

``` python
class Settings(HasTraits):
    a = Int(1)
    b = List(Str('foo'), Str('bar'))
```
One problem with this approach is that the namespace of the ``HasTraits`` class is hugely cluttered so all the actual settings get buried/lost.  This can be remedied by making the ``HasTraits`` class a hidden attr of ``Settings`` (i.e. ``Settings._settings``) and overloading the setattr and getattr methods to look in the ``_settings`` attr.  This is a bit complicated but works ok.

The ``attr`` package offers the benefits of no cruft, but the data protections are much more complicated:

``` python
@attr.s
class Settings:
    a = attr.ib(validator=attr.validators.instance_of(int),
                on_setattr=attr.setters.validate)
```

* Limit the cruft in the settings name space.  We don't want the settings attribute to be a dict or have any methods....just the set attributes.  

* The ability to inherit settings between subclasses of algorithms should be easy and clean.  This is where things fall apart.  Each algorithm file containing their respective settings class and their actually algorithm class.  To inherit from the settings class of a parent we need to import it. This will pollute the name space of the algorithms module.  Note that the current approach, explained in issues #2108 avoid this but is a bit awkward.  It converts the settings class into a dict and adds the values to a dict on the algorithm so at each level of inheritance the dict fills up.  This requires some hacking to make the docstring thing work.  It's complicated and hence feels awkward.


jgostick:
Fixes #1951 
Fixes #2108
Fixes #2077

PR #2128, opened 2021-11-03
Clean up network classes

jgostick:


jgostick:
this was opened in error

PR #2159, opened 2021-11-12
Organize examples folder to match porespy repo

### LINKED ISSUE #2008: Reorganize examples folder to match the porespy repo
* Create examples.rst file to generate the "Examples" tab in the homepage
* Copy examples_linker.py into the docs folder, etc
* Reorganize the examples as:
  * Move algorithms/* to an "Examples" folder
  * Move the rest of the folders (i.e. geometry, networks) to a "Tutorials" folder
  * Notebooks already in the Tutorials folder should be moved to subfolder called "Usage Guides" or "General Usage"
  * Perhaps these could be divided as "Overviews" and "Focused"
* Paper recreations to the "Contrib" folder
* Leave the module reference part alone...that will be handled in a separate issue

### LINKED ISSUE #2348: Fix submodule imports
VERY IMPORTANT: apply the following changes directly to the `organize_examples` branch!

---

Suppose we have the following directory structure:

```
openpnm
	module_A
		_funcs.py
		__init__.py
	module_B.py
	__init__.py
```

where the functions/classes for `module_A` are dumped inside `_funcs.py`, whereas those for `module_B` are dumped inside `module_B.py`.

with `openpnm/__init__.py` being:

```
from . import module A
from . import module B
```

and `openpnm/module_A/__init__.py` being:

```
from ._funcs import *
```

The ``__all__`` trick works fine for `module_A` (since it's a separate folder with its own `__init__.py` file containing a * import), whereas it fails for `module_B`. To fix it, we need to turn `module_B.py` into its own folder and `__init__.py` file.

We have this problem in `openpnm` in a few places. It'd be great if someone could volunteer to fix these:

- [x] `openpnm.models.phase`
- [x] `openpnm.models.physics`

Here's an example on how to do it: In `openpnm.models.phase`, we have `density.py`. So we need to:

- Create a new folder: `openpnm/models/phase/density`,
- Rename `density.py` to `_funcs.py`
- move `_funcs.py` to the newly-created-folder
- Create `__init__.py` with the content: `from _funcs import *` and place it inside the newly-created folder
- Apply the `__all__` trick to `_funcs.py`, meaning that if it contains `method_A` and `method_B`, add the following line to the beginning of `_funcs.py` (right after the imports with double blank lines): `__all__ = ["method_A", "method_B"]`

mkaguer:
Fixes #2008 
Fixes #2348


mkaguer:
The Examples README file also needs updating based on new file structure. Links are broken.

PR #2192, opened 2021-11-25
Update repr on Base to include name

### LINKED ISSUE #2100: Update repr on Base to include name
A simple change but would make the repr much more useful:

``` python
def __repr__(self):
    return f'{self.name} <{self.__class__.__module__} object at {hex(id(self))}>'
```

hfathian:
Fixes #2100


ma-sadeghi:
@hfathian I think it still needs some work, here's the output that I'm getting:
```python
>>> net = op.network.Cubic([1,2,3])
>>> net
<(<class 'openpnm.network.Cubic.Cubic'>, <class 'openpnm.network.GenericNetwork.GenericNetwork'>, <class 'openpnm.core.Mixins.ParamMixin'>, <class 'openpnm.core.Base.Base'>, <class 'dict'>, <class 'openpnm.core.ModelsMixin.ModelsMixin'>, <class 'openpnm.core.Mixins.LabelMixin'>, <class 'object'>) object at 0x7f83229b7c70>
```

PR #2237, opened 2021-12-15
Rename phases to phase

### LINKED ISSUE #2058: Rename phases to phase

mkaguer:
fixes #2058


mkaguer:
There are cases where phases refers to a list of phase arguments. See _multiphase.py for example. I did not change these from phases to phase but should I? Also, see io folder, similar case there.

mkaguer:
left project.phases() alone. Also, renamed models/phases to phase.

PR #2244, opened 2021-12-15
Changing regenerate model to work with models without 'regen_mode' key

### LINKED ISSUE #2235: Make regen_mode optional
If not found the it should be assumed as normal, and maybe also regen mode could be added to models dict.  It would be nice for the new models.collections to not have to specify regen mode.

Ni2M:
This PR closes #2235. 
If the model for any reason does not have 'regen_mode' in its keys, the regenerate_models counts it as a regen='normal' and runs the model to update its associated property. If the regen_mode is already in the model keys (which by default is 'normal'), the regenerate_models follows that already defined mode.
An example where 'regen_mode' is not in model key would be the new collection of models. But we can recreate the same example problem:
``` python
import numpy as np
import openpnm as op
np.random.seed(10)
shape=[10, 10, 10]
pn = op.network.Cubic(shape)
geo = op.geometry.SpheresAndCylinders(network=pn, pores=pn.Ps, throats=pn.Ts)
geo.models['pore.diameter'].pop('regen_mode')
geo.regenerate_models()
#KeyError: 'regen_mode'
```
Fixed:
``` python
import numpy as np
import openpnm as op
np.random.seed(10)
shape=[10, 10, 10]
pn = op.network.Cubic(shape)
geo = op.geometry.SpheresAndCylinders(network=pn, pores=pn.Ps, throats=pn.Ts)
geo.models['pore.diameter'].pop('regen_mode')
print(geo['pore.diameter'].mean())
# 0.4460026112474829
geo['pore.seed']=0.5
geo.regenerate_models()
print(geo['pore.diameter'].mean())
# 0.4525925641219549
```

### DIFF
--- a/openpnm/core/_models.py
+++ b/openpnm/core/_models.py
@@ -346,8 +346,11 @@ class ModelsMixin:
         if propnames is None:  # If no props given, then regenerate them all
             propnames = self.models.dependency_list()
             # If some props are to be excluded, remove them from list
-            exclude.extend([k for k, v in self.models.items()
-                            if v['regen_mode'] == 'explicit'])
+            for k, v in self.models.items():
+                if 'regen_mode' not in v:
+                    pass
+                elif v['regen_mode'] == 'explicit':
+                    exclude.extend([k])
             propnames = [i for i in propnames if i not in exclude]
         # Re-order given propnames according to dependency tree
         self_models = self.models.dependency_list()


Ni2M:
Note to myself: We should make sure this works upon **adding a model** without regen key and not just upon **calling regenerate_models** (like similar to collection models).

Ni2M:
I checked the concern for upon adding a model, and upon adding it will by default add regen_mode='normal' [here] (https://github.com/PMEAL/OpenPNM/blob/635e2539a1ef92e5e59bf1d19af53222893c86a1/openpnm/core/_models.py#L306). So it should be fine for those cases.

PR #2253, opened 2021-12-16
Enhance _validate_data_health to find throats without an assigned conductance model

### LINKED ISSUE #1678: Enhance GenericTransport._validate_data_health to account for locations w/o explicit models
Basically, in multidomain problems, if you forget to assign a model to one of the subdomains, your conductance array will contain nans in some locations, and if you run an algorithm, the health check is clueless, but this particular case can be identified.

ma-sadeghi:
Fixes #1678

### DIFF
--- a/openpnm/algorithms/_generic_transport.py
+++ b/openpnm/algorithms/_generic_transport.py
@@ -218,18 +218,17 @@ class GenericTransport(GenericAlgorithm, BCsMixin):
         in the ``settings`` attribute.
 
         """
-        # Initialize the solution object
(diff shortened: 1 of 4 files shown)


jgostick:
I'm not sure I love this.  I often set conductance values by hand when testing / prototyping.  Won't this raise an exception?

ma-sadeghi:
@jgostick It only gets triggered if `nans` are found in the A matrix. `_validate_data_health` gets short-circuited if A and b are sane.

PR #2256, opened 2021-12-16
Add shale conductance example for version 2.8.0

xu-kai-xu:
@jgostick @ma-sadeghi , sorry for my rushed PR for this example. Now I have cleaned others and only keep the example for version 2.8.0.


ma-sadeghi:
@xu-kai-xu Thanks. I think I can still see some v2.5 files under "Files Changed". Could you please take a look?

PR #2260, opened 2021-12-16
Add MultiPhysics class

ma-sadeghi:


openpnm/algorithms/_transient_multiphysics.py
```diff
@@ -0,0 +1,121 @@
...
+
+    Parameters
+    ----------
+    %(GenericAlgorithmSettings.parameters)s
+    algorithms: list
+        List of transient algorithm objects to be solved in a coupled manner
+
+    """
+    algorithms = []
+
```
ma-sadeghi: @mkaguer You need another blank like here. Basically, PEP8 suggests two blank lines between methods/classes in a module

PR #2270, opened 2021-12-20
Fix _xdmf file to show pore coords in Paraview

### LINKED ISSUE #2123: Reversed dimensions in XDMF output lead to unreadable data for ParaView
The XDMF output creates HDF5 files with dimensions given by `ndarray.shape`:
https://github.com/PMEAL/OpenPNM/blob/f176ea2cd76f7879edc7eabe13115a6ba9a8b18e/openpnm/io/XDMF.py#L109-L118
However, in the XDMF file, the dimensions are reversed:
https://github.com/PMEAL/OpenPNM/blob/f176ea2cd76f7879edc7eabe13115a6ba9a8b18e/openpnm/io/XDMF.py#L130-L131
https://github.com/PMEAL/OpenPNM/blob/f176ea2cd76f7879edc7eabe13115a6ba9a8b18e/openpnm/io/XDMF.py#L139-L143
This results in unreadable data for [ParaView](https://www.paraview.org/), which replaces it by zeros.
This is fixed, when replacing [Line 131](https://github.com/PMEAL/OpenPNM/blob/f176ea2cd76f7879edc7eabe13115a6ba9a8b18e/openpnm/io/XDMF.py#L131) by
```.py
dims = (' '.join([str(i) for i in shape]))
```

Furthermore, there is a small typo:
https://github.com/PMEAL/OpenPNM/blob/f176ea2cd76f7879edc7eabe13115a6ba9a8b18e/openpnm/io/XDMF.py#L223
This should be:
```.py
'DataType': "Float",
```

Ni2M:
This PR closes #2123. The network was shown in correct xyz order in Paraveiw. The problem was the pore.coords values were all zero. 
Sample code for testing:
```python
import openpnm  as op
import numpy as np
np.random.seed(5)
net = op.network.Cubic(shape=[2, 2, 2])
geo=op.geometry.SpheresAndCylinders(network=net, pores=net.Ps, throats=net.Ts)
op.io.to_xdmf(network=net, filename='test')
```
Problem in pore coords:
image

Solved:
image

### DIFF
--- a/example.py
+++ b/example.py
@@ -2,7 +2,7 @@ import numpy as np
 import openpnm as op
 from openpnm.models.physics import source_terms
 
-# %% Initialization: create Workspace and project objects.
+# %% Initialization: Create Workspace and project objects.
 ws = op.Workspace()
 ws.settings.loglevel = 50
 np.random.seed(9)
--- a/openpnm/io/_xdmf.py
+++ b/openpnm/io/_xdmf.py
@@ -127,7 +127,8 @@ class XDMF(GenericIO):
                     ):
                         attr_type = 'Scalar'
                         shape = D[item].shape
-                        dims = (''.join([str(i) + ' ' for i in list(shape)[::-1]]))
+                        #dims = (''.join([str(i) + ' ' for i in list(shape)[::-1]]))
+                        dims = (' '.join([str(i) for i in shape]))
                         if '@' in item:
                             item = item.split('@')[0]+'@t'
                             hdf_loc = fname_hdf + ":" + item
--- a/openpnm/io/_xdmf.py
+++ b/openpnm/io/_xdmf.py
@@ -127,7 +127,6 @@ class XDMF(GenericIO):
                     ):
                         attr_type = 'Scalar'
                         shape = D[item].shape
-                        #dims = (''.join([str(i) + ' ' for i in list(shape)[::-1]]))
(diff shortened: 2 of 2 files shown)


Ni2M:
This PR works on my local device. However here the build test gives an error related to scikit-fmm.

ma-sadeghi:
@Ni2M It seems that the problem only occurs with Python 3.9 on Windows. I'd say let's skip that particular test on Python 3.9 on Windows for now, until it gets resolved. You can decorate the failing test like the following::

```python
import sys
import pytest

condition = sys.version_info >= (3, 9) and sys.platform.startswith("win")
@pytest.mark.skipif(condition, reason="scikit-fmm is broken on Windows + Python 3.9")
def failing_test(): ...
```

Ni2M:
Thank you @ma-sadeghi for your help and investigation. Currently the pip install fails, which I think it is related to the requirements. However, I couldn't find the sk-fmm in the requirements list.  As it is not a specific testfile or .py file that is failing, I was wondering if you could please let me know where should I include the script to skip the test. Thanks.

ma-sadeghi:
Right, I thought some tests were failing. So I'd say let's merge this PR. It seems that skfmm devs are [working](https://github.com/scikit-fmm/scikit-fmm/issues/64) on python 3.9 wheels, so should be fixed soon.

Issue #2297, opened 2022-01-18
Remove "results" method from algorithms?


jgostick:
At the moment you have to do ``phase.update(alg.results())`` to get the "results" put back on the phase object. The reason we did this was to avoid 'polluting' the phase object without the user's permission.  However, this is a bit pointless since any iterative props and source terms already update the physics and phase objects during solution.  So it might make sense to just do it automatically at the end of the solution process.

jgostick:
BUT, we have now added the ``Solution`` object that gets returned by ``run`` methods.  SOOO, instead of using ``result`` the user can just do ``phase.update(alg.soln)`` or ``phase.update(alg.run())``.  Note that transient algs return a ``TransientSolution`` object, which carries with it the ``interpolate`` method and the ``t`` attribute.  This can be used by the IO classes to create the necessary time stamps.  

Also, for the percolation algorithms, if we change the run methods to only do ONE point, then the use would tell the alg which pressure or volume to apply, then it would result the resulting configuration. So ``phase.update(op.run(pressure=2000))`` would return the invasion configuration at an applied pressure of 2000 Pa.

jgostick:
We should add some method to ``Solution`` that returns a dictionary of values with keys that contain the @ timestamp notation.  This way the user could generate the results at the times they desire, as follows:

``` python
>>> dict = alg.soln.to_dict(propname="pore.concentration", timestamps=[0.1, 2, 3, 4])
OR
>>> print(dict)
{'pore.concentration@0.1': ndarray, 'pore.concentration@2': ndarray, ...}
```

Then they could update the phase with this dict:

``` python
phase.update(dict)
```

Alternatively, they could just put the entire ``Solution`` array onto the phase object with `phase.update(alg.soln)`.  `alg.soln.t` is time stamps where an actual solution was obtained, so the IO class, if it detects a 'transient' solution on the give phase, would use these.

jgostick:
Summary:
1. Yes, we should remove the ``results`` method from all algorithms
2. The ``soln`` object is now already stored in a dict, so the ``phase.update(alg.soln)`` already works.  
3. The @ symbol notation is going to be dropped in V3, since now all transient data is in the soln object.  How this impacts the paraview output remains to be solved.
4. The percolation algorithms will be handled like transient ones, with pressure steps instead of time steps.

Issue #2296, opened 2022-01-12
Metrics 'run' methods need to pass args to the algorithms


jgostick:
I'm having trouble using pardiso at the moment so need to specify algs to use the scipy solver.  This works fine with the run method on all the main algs, but metrics call run within their own run.  We should pass kwargs through to the internal run method.

Ni2M:
During the dev meeting we have decided to include the default solver in the workspace settings. A PR will be created to address this issue.

PR #2301, opened 2022-01-19
Fixing metrics issues

### LINKED ISSUE #2298: Remove use of results from metrics classes

Ni2M:
This PR closes #2298, #2296, #2297


Ni2M:
We have added the default solver to the workspace settings. However, we did not revoke the ability to pass the solver as an argument to the generic transport alg and its subclasses.
https://github.com/PMEAL/OpenPNM/blob/b60786a2194e9812ef3728b76f75d483aebd6b38/openpnm/algorithms/_generic_transport.py#L204

PR #2333, opened 2022-02-24
Replace pore surface area for loop with numpy unbuffered operation

### LINKED ISSUE #2327: pore_area models are needlessly slow
We're using list comprehensions in these models, but could use unbuffered vector operations:

``` python

import numpy as np
import openpnm as op

net = op.network.Cubic(shape=[3, 3])

net['pore.diameter'] = 1
net['throat.diameter'] = np.random.rand(net.Nt)*0.5
net['throat.length'] = 1

net['throat.cross_sectional_area'] =  4/3*np.pi*net['throat.diameter']**2
net['throat.surface_area'] = np.pi*net['throat.diameter']*net['throat.length']
# Find surface are of sphere
temp = 4*np.pi*(0.5*net['pore.diameter'])**2
# Subtract throat openings
a = net['throat.cross_sectional_area']
np.subtract.at(temp, net.conns.flatten(), np.hstack((a, a)))
# Readd throat surface areeas
a = net['throat.surface_area']/2
np.add.at(temp, net.conns.flatten(), np.hstack((a, a)))

```

Ni2M:
This PR closes #2327. We use [numpy.subtract.at](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.at.html) to perform in place subtraction.
- [x] Update surface area functions
- [x] Tests check
- [x] Check for multi-geom networks

### DIFF
--- a/openpnm/models/geometry/pore_surface_area/_funcs.py
+++ b/openpnm/models/geometry/pore_surface_area/_funcs.py
@@ -73,8 +73,7 @@ def circle(
 
     """
     network = target.project.network
-    R = target[pore_diameter] / 2
-    value = 2 * _np.pi * R
+    value = _np.pi * target[pore_diameter]
     Tca = network[throat_cross_sectional_area]
     _np.subtract.at(value, network.conns.flatten(), _np.repeat(Tca, repeats=2))
     return value
@@ -100,7 +99,7 @@ def cube(
     """
     network = target.project.network
     D = target[pore_diameter]
-    value = 6 * D**2
+    value = 6.0 * D**2
     Tca = network[throat_cross_sectional_area]
     _np.subtract.at(value, network.conns.flatten(), _np.repeat(Tca, repeats=2))
     return value
@@ -126,7 +125,7 @@ def square(
     """
     network = target.project.network
     D = target[pore_diameter]
-    value = 4 * D
+    value = 4.0 * D
     Tca = network[throat_cross_sectional_area]
     _np.subtract.at(value, network.conns.flatten(), _np.repeat(Tca, repeats=2))
     return value
--- a/openpnm/models/geometry/pore_surface_area/_funcs.py
+++ b/openpnm/models/geometry/pore_surface_area/_funcs.py
@@ -46,6 +46,7 @@ def sphere(
     value = 4 * _np.pi * R**2
     Tca = network[throat_cross_sectional_area]
     _np.subtract.at(value, network.conns.flatten(), _np.repeat(Tca, repeats=2))
+    value = value[network.pores(target.name)]
     return value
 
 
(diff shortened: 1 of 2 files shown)


Ni2M:
I realized because the indices of pores and throats in each geom is based on 0,1, .. and not their indice in the network pores, we can't relate the values of surface area calculated from all conns to the target pores. Therefore, we may need to find another way of calculating them (not based on net.conns). Here is some examples that failed:
image


```python
import openpnm as op
import openpnm.models.geometry.pore_surface_area as mods
net = op.network.Cubic(shape=[10, 1, 1], spacing=1.0)
geo1 = op.geometry.GenericGeometry(network=net,
                                        pores=net.Ps[0:3],
                                        throats=net.Ts[0:3])
geo2 = op.geometry.GenericGeometry(network=net,
                                        pores=net.Ps[3:],
                                        throats=net.Ts[3:])
print(geo1.Ps, geo1.Ts) # gives us [0,1,2] [0,1,2]
print(geo2.Ps, geo2.Ts) # gives us [0 1 2 3 4 5 6] [0 1 2 3 4 5]
# does not give us [3,4,5,6,7,8,9] [3,4,5,6,7,8] to use them for returning the pores SA based on network indices
geo1['pore.diameter'] = 1
geo1['throat.cross_sectional_area'] = 0.3
geo2['pore.diameter'] = 1
geo2['throat.cross_sectional_area'] = 0.1
geo2.add_model(propname='pore.surface_area',
                    model=mods.circle,
                    regen_mode='normal')
geo1.add_model(propname='pore.surface_area',
                    model=mods.circle,
                    regen_mode='normal')
```

model: 
```python
# model:
def circle(
    target,
    pore_diameter='pore.diameter',
    throat_cross_sectional_area='throat.cross_sectional_area'
):

    network = target.project.network
    value = _np.pi * network[pore_diameter]
    Tca = network[throat_cross_sectional_area]
    _np.subtract.at(value, network.conns.flatten(), _np.repeat(Tca, repeats=2))
    target_pores = target.Ps
    # this does not give us correct results because target.Ps is based
    #on geom indices starting from 0 not the net indices
    value = value [target_pores]
    return value
   ```
   
   one solution may be using:
   ```python
    target_pores = _np.unique(network.conns[target.Ts])
 ```
  
but still it is not correct, as target.Ts is based on  geom indices that starts from 0 not the net indices for geom2 target_pores=[0 1 2 3 4 5 6] while it should be : [3,4,5,6,7,8,9] . Another problem with this is a throat may have  one pore in geom1 and another one in geom2, so getting the  pores from throat conns would not be correct even if the indices 
of the throats were based on net indices.

Ni2M:
I've updated the code and the "surface_area" branch is working now. For some reason I don't see my final commit here in the PR, but it has passed the tests on surface_area branch see [link for code](https://github.com/PMEAL/OpenPNM/commit/da8ea1e242c9892100db3cd14aec9ddc89b45dc4) and [link for tests](https://github.com/PMEAL/OpenPNM/actions/runs/2023425558). That surface_area branch is ready to merge, if it is still connected to this PR.

Ni2M:
The problem with the commit is solved. This PR is ready to merge.

PR #2384, opened 2022-04-22
Refactoring percolation algs ongoing

jgostick:
The previous version of this PR had some important changes on it, so we merged it.  This PR will continue the work of updating the percolation algorithms that was left unfinished.


jgostick:
These failing tests make no sense.  Everything passes locally, I even updated my numba and networkx to the versions used here, but do not reproduce any of these failures.  

This is another case where it would be nice if the CI build outputed the 'conda freeze' file so we could reproduce the *exact* environment.  When the tests all pass we could put that file on the release, while if the tests fail we could debug.

PR #2420, opened 2022-05-10
Create plot_network.py

ivonindima:
Adds a script with an example of three-dimensional visualization of the network using VisPy.


jgostick:
I'm going to merge this function into the ``contrib`` folder, then we can move it into plottools in the near future. Thanks very much @ivonindima !

PR #2444, opened 2022-05-18
Update one error message in _generic_transport.py

mzdreambig:
Since the "trim_pores" is not an attribute of network health.


ma-sadeghi:
@mzdreambig Thank you so much for your contribution!

PR #2470, opened 2022-05-27
Improved settings

jgostick:


ma-sadeghi:
@jgostick Just asking: what was wrong with the dataclass?

ma-sadeghi:
@jgostick One more question: Can we still enforce specific types? Like source terms as sets, etc.

jgostick:
@ma-sadeghi 
It is a 'dataclass-like' object.  The python dataclass does not do anything useful actually, so we just wrote our own.  And it still should enforce the data type.  All I've changed here is that we no longer accept ``settings`` in the init.  If you want to change the settings, you must do it after initialization.  The process of passing the settings up the tree using ``super().__init__(settings)`` was complicated and only necessary because I had the "bright" idea of putting the name and the prefix for names in the settings.  This created a very large burden to support an unnecessarily cute idea.  So now, the name is accepted as an argument in init, like we've always done.  The devs can put a default name in the init, like ``Phase(name='air_#')``, where the hashtag indicates to the name checker that is should be replaced with the next available number, OR the user can put a name like ``Phase(name='bob')``, which will just override the default value.  Aside from the # trick, it's all very simple and transparent.

PR #2474, opened 2022-05-31
Fixed exporting transient data to paraview

### LINKED ISSUE #2068: Update Paraview output to work with new Solution class format
Paraview needs transient data from each time step in its own file, so the new Solution class needs to output this somehow.

mkaguer:
Fixes #2068 

I want to see if this breaks any unit tests. But I think I got it working. When `project_to_dict` is called it takes transient solution data from algorithm object now and adds to dictionary. The @ symbol is added for any transient data. Thereafter, the rest of the code works.

### DIFF
--- a/openpnm/io/_dict.py
+++ b/openpnm/io/_dict.py
@@ -1,5 +1,6 @@
 import logging
 from openpnm.utils import NestedDict
+from openpnm.utils._misc import _is_transient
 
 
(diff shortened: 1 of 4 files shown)


mkaguer:
Hey @jgostick. Is `project_to_paraview` supposed to work with transient data? It seems like a different beast!

PR #2495, opened 2022-07-03
Refining metrics

### LINKED ISSUE #2496: Should metrics return a project?
At the moment they are basically treated as algorithms.  The downsides of this are:

* The phases (i.e. Hg) and algorithm are added to the project, which clutters things. 
* The metric object has a bunch of attributes on it, like ``pores`` and ``num_throats``. These are not really needed, and they only hide the key methods that are added by the metrics, like ``plot_pc_curve``.

Instead, I wonder if the metrics should return a new ``Project`` which contains a copy of the network, all the phases, and the algorithms.  This project could have all the needed methods and attributes attached to it.  Also, for this idea I have pushed a PR with a Project class that has the bare minimum of methods, making room for the specialized attributes.

jgostick:
Playing with the idea that ``metrics`` should return a ``Project`` object.

Fixes #2496


jgostick:
We are closing this in favor of removing this functionality altogether, and instead using the code to beef up our example notebooks.

PR #2543, opened 2022-08-11
Modifying conduit models

### LINKED ISSUE #1891: Ability to handle diverging cones?
As discussed in [this issue](https://github.com/PMEAL/porespy/issues/97) on PoreSpy, it can occur that throats are larger than the two pores which they join.  This is not necessarily unphysical.  ]

We need to ensure that openpnm's new conductance models (see PR #1604) can handle 'diverging' cones.  The ``cones_and_cylinders`` model is probably the place for this.

### LINKED ISSUE #1960: Add intersecting cones and pyramid models
It might be handy to have pore-scale conductance models with no throats...just two cones or pyramids intersecting at the throat centroid.

### LINKED ISSUE #2363: Size factor models return negative numbers for conduits with "engulfing" pores
When one pore center lies within another pore, our size factor models break down, including `cones_and_cylinders`.

### LINKED ISSUE #2365: Fine-tuning conductance models
Extracted networks occasionally have issues with pores that overlap too much (see #2363).  Too much means the center of one pore is beyond the lens of the other pore.  

There are two solutions we should implement:

1.  Add 2 new conductance models: Intersecting cones and intersecting pyramids (see #1960).  These models require a 'throat.centroid' which we currently output in ``snow2`` as ``throat.global_peak``.  So the models would have an argument like ``throat_centroid='throat.global_peak'``.  We should probably export ``throat.centroid`` from ``snow2`` in the near future.
2. Update the existing models to throw a hard exception if pores overlap too much...instead of trying to do an 'opinionated' correct on the fly.

Ni2M:
Closes #2365
closes #2363 
closes #1960 
closes #1891 

Note for dev meeting: see ppt.
Tasks:
conduit length: 
- [x] hybrid, intersectiong cones and cylinders. 
- [x] Pyramids are wrappers (added wrapper methods)
- [x] trapezoids are wrappers (added wrapper methods)

Throat length:
- [x] wrappers for all new models

diff_size factors: 
- [x] hybrid, intersectig cones and cylinders
- [x] hybrid, intersecting pyramids and cuboids
- [x] hybrid, intersecting trapezoids

Hyd_size factors: 
- [x] hybrid, intersectig cones and cylinders
- [x] hybrid, intersecting pyramids and cuboids
- [x] hybrid, intersecting trapezoids

Tests:
- [x] unit tests for intersecting and hybrid models
- [x] drafted local example to test extracted net, conductance
- [x] final check on array length, questions in ppt


Ni2M:
Draft of thoughts, example:
[modify cond models.pptx](https://github.com/PMEAL/OpenPNM/files/9338937/modify.cond.models.pptx)
Note: Other shapes and size factors follow a similar process. They will be updated based on current changes and any suggestions/feedbacks.

Ni2M:
Notes: (details in ppt)
- The size factor formulations for cones, pyramids and trapezoids already handle throats diam larger than pores diam
- Throat.spacing in conduit length is kept with a try except for CTC calcs, because not all nets have throat.spacing (like extracted nets).
- Default throat_coords is throat.coords. The user can define it before calling the function either by calling throat_centroid model in geom or assigning throat.global_peak (if existed) to throat.coords.
- Notices the argument for models: In diff_size_factor they were all based on target, in hyd_size factor they were all based on network.

Ni2M:
@jgostick Final changes are done. Thanks.

PR #2564, opened 2022-09-07
Improve code coverage

ma-sadeghi:


ma-sadeghi:
The coverage after this PR is at 81.4% from ~70%, so I'm merging it. We can add more tests later.

PR #2648, opened 2022-12-15
Moving import of h5py inside function

jgostick:
h5py does not work with python 3.10 at the moment.  I am moving their import inside their respective functions so that at least openpnm can be imported.  No sure what flake8 will say...


ma-sadeghi:
@jgostick Is this a Windows issue? because I can import it on Ubuntu.

jgostick:
I included this in the recent PR #2672

PR #2664, opened 2023-01-05
Fixing delaunay and voronoi generators

### LINKED ISSUE #1895: Add lloyd_relaxation arg to Voronoi and Delaunay tellesllations
https://en.wikipedia.org/wiki/Lloyd%27s_algorithm

### LINKED ISSUE #2635: `reflect_base_points` should remove point lying outside the given domain first
I was playing with Lloyd relaxation (#1895) and found that I had to manually trim base points that lay outside the domain on each iteration.  Otherwise these points are reflected during the call to ``reflect_base_points`` and resulted in a mess.

### LINKED ISSUE #2646: `DelaunyVoronoiDual` (and maybe `Voronoi`) break when list of 2D points given
There is a line in there that is supposed to be ok with 2D points, but fails.

jgostick:
Fixes #2646 
Fixes #2635 
Closes #1895


jgostick:
There is one test failing, which is caused by the Dijkstra algorithm returning something different in V1.10 than in previous versions.  I have opened a [bug report](https://github.com/scipy/scipy/issues/17734) on the scipy repo.

jgostick:
Mac builds are failing again, but this seems to be a common thing so I'm going to ignore it for now and merge this.

PR #2672, opened 2023-01-19
Fix io bug

### LINKED ISSUE #2662: proj_to_vtk  breaks when drainage and/or invasion algorithms present
As mentioned in #2659 the io function is looking for a soln attribute on the algorithms and not finding it.  Also I think it's looking for a quantity key in the settings.  Basically the io function is assuming that only transport algorithms are present.  

Actually, I'm wondering why it's looking for algorithms at all...I think these functions only output the network and phases?

jgostick:
Fixes #2662


jgostick:
mac builds failed again.  Didn't even build.  Ignoring.

PR #2761, opened 2023-06-07
Checking isfinite in plot methods

jgostick:


jgostick:
This is a minor detail, but because we normalize the size_by and color_by inputs by their maximums, the sizes/colors were all being set to 0.  Took me half an hour to figure out why, so thought I'd add this check.

PR #2781, opened 2023-06-28
Add thermal conduction

AlemSnyder:
I though it was odd that there is no transient Fourier conduction.


jgostick:
Hi @AlemSnyder 
Thansk for this, and sorry for my late response. It slipped off my radar, so this new commit was a good nudge.  I will reveiw it soon.

AlemSnyder:
Sorry, that commit was an accident. I needed it for something else.

I reverted the branch to the correct state.

jgostick:
I noticed that the second commit was a bit out of scope, but seems like a good idea to have anyway.  Feel free to push that as a separate commit.

PR #2798, opened 2023-08-09
Fix ndarray dimension reduction in `Transport.rate`

### LINKED ISSUE #2796: Traceback when getting flow rate of a network with a single throat
Running the following line:
`perm.rate(throats=perm.network.throats("all"), mode="individual")`
(where perm is a openpnm.algorithms.StokesFlow of a network with a single throat)
throws the following error:
```
  File "C:\Work\lib\Python\Lib\site-packages\openpnm\algorithms\_transport.py", line 318, in rate
    R = np.absolute(Qt[throats])
IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```

This seems to be caused by the following line squeezing a ndarray with more than one dimension, but a single value, into a zero-dimension array when it has a single value, while the rest of the code still treats the variable as a one dimension array:
https://github.com/PMEAL/OpenPNM/blob/0efb343f2f564b4cf0189828d49096da1addddda/openpnm/algorithms/_transport.py#L315C46-L315C46
`        Qt = np.diff(g*X12, axis=1).squeeze()`
        
I fixed the problem locally with the following code below the previous line:
```
        if Qt.ndim == 0:
            Qt = Qt[np.newaxis]
```
   
Is this solution acceptable? Can I make a pull request to fix it in the repository?

Arenhart:
Fixes #2796

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 name: Tests
 
-on: [push]
+on: [pull_request]
 
 jobs:
   build:
@@ -15,7 +15,6 @@ jobs:
       max-parallel: 9
       matrix:
         python-version: ['3.9', '3.10', '3.11']
-        # Add macos-latest to the next line once #2451 is fixed
         os: [ubuntu-latest, windows-latest, macos-latest]
         include:
         - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ma-sadeghi:
Looks good! I'll merge once the tests pass. Thank you @Arenhart once again!

PR #2810, opened 2023-08-31
New release (patch): v3.2.1

jgostick:
There have been a few useful additions and important fixes to the dev branch that we should release.


jgostick:
I think this should be a patch version, since the new stuff is so minor that it's not really worth writing a release notes for.

PR #2853, opened 2023-11-03
More robust conduit lengths

jgostick:
The `spheres_and_cylinders` conduit length model was not handling an edge case properly.  When two spheres overlapped, AND the throat was larger than the area of overlap, it was ignoring the contribution of the throat. It was basically assuming that the throat was much smaller than the pores.

I also added an extra check to this function so it raises an exception if the spheres overlap TOO much. We have had some people noticing negative flows and other odd behavior if they weren't being careful about spacing and pore sizes, so I decided to add this check even though we wanted to avoid too much "hand holding".  I think this check is general enough that its acceptable.

### DIFF
--- a/openpnm/models/geometry/conduit_lengths/_funcs.py
+++ b/openpnm/models/geometry/conduit_lengths/_funcs.py
@@ -51,15 +51,31 @@ def spheres_and_cylinders(
     # Handle the case where Dt > Dp
     if (Dt > D1).any() or (Dt > D2).any():
         _raise_incompatible_data()
+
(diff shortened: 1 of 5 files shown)


ma-sadeghi:
@jgostick Should we instead add a health check to the `run` method? The very least, check if the conduit array (Nt by 3) is all positive.

jgostick:
I think it'd be a good idea to have a robust "check_physics_health" function that reports on errors.  I'd rather not call it on *each* run though, since it slows things.

PR #2857, opened 2023-11-16
Creating a patch release to fix matplotlib import error

jgostick:
The deprecation of the `matplotlib.docstring` module to `matplotlib._docstring` was fixed in our previous patch, BUT it was not compatible with both new and old versions of matplotlib, so this patch uses a try/except to work with both versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,40 +10,41 @@
 
 # Overview of OpenPNM
 
-*OpenPNM* is a comprehensive framework for performing pore network simulations of porous materials.
+OpenPNM is a comprehensive framework for performing pore network simulations of porous materials.
 
 ## More Information
 
-For more details about the package can be found in the [on-line documentation](https://openpnm.org)
+For more details about the package can be found in the [online documentation](https://openpnm.org)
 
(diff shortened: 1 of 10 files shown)


ma-sadeghi:
@jgostick Given that Matplotlib is [going to remove this module](https://matplotlib.org/stable/api/docstring_api.html) altogether, can't we just host the code in our side? I think the license allows it.

jgostick:
The windows builds are failing because the pypardiso team is having trouble with mkl again.  They have a lot of recent activity and I'm sure they'll get this fixed soon.  I am tempted to merge this now so we can get the pip release working with all version of mpl, which is more important than pardiso, I think?

ma-sadeghi:
Maybe give them just a few more days? Either the issue is on pypardiso side, which I think will get fixed soon, or on Intel side, which probably won't get fixed soonish

jgostick:
But technically pypardiso is not in our dependencies.  This is only failing because we use it in some of our tests...but you're right, lets give it a few more days.

Issue #2863, opened 2023-11-20
Matplotlib is breaking our use of docstring.substitution


jgostick:
I have fixed this by doing a try/except on the import of the function, but we need to remove this completely because I'm sure they'll break it again.

ma-sadeghi:
As a quick fix, we can include Matplotlib's `_docstrings` private module in OpenPNM, it's quite light-weight:

```python
import inspect

from . import _api


def kwarg_doc(text):
    """
    Decorator for defining the kwdoc documentation of artist properties.

    This decorator can be applied to artist property setter methods.
    The given text is stored in a private attribute ``_kwarg_doc`` on
    the method.  It is used to overwrite auto-generated documentation
    in the *kwdoc list* for artists. The kwdoc list is used to document
    ``**kwargs`` when they are properties of an artist. See e.g. the
    ``**kwargs`` section in `.Axes.text`.

    The text should contain the supported types, as well as the default
    value if applicable, e.g.:

        @_docstring.kwarg_doc("bool, default: :rc:`text.usetex`")
        def set_usetex(self, usetex):

    See Also
    --------
    matplotlib.artist.kwdoc

    """
    def decorator(func):
        func._kwarg_doc = text
        return func
    return decorator


class Substitution:
    """
    A decorator that performs %-substitution on an object's docstring.

    This decorator should be robust even if ``obj.__doc__`` is None (for
    example, if -OO was passed to the interpreter).

    Usage: construct a docstring.Substitution with a sequence or dictionary
    suitable for performing substitution; then decorate a suitable function
    with the constructed object, e.g.::

        sub_author_name = Substitution(author='Jason')

        @sub_author_name
        def some_function(x):
            "%(author)s wrote this function"

        # note that some_function.__doc__ is now "Jason wrote this function"

    One can also use positional arguments::

        sub_first_last_names = Substitution('Edgar Allen', 'Poe')

        @sub_first_last_names
        def some_function(x):
            "%s %s wrote the Raven"
    """
    def __init__(self, *args, **kwargs):
        if args and kwargs:
            raise TypeError("Only positional or keyword args are allowed")
        self.params = args or kwargs

    def __call__(self, func):
        if func.__doc__:
            func.__doc__ = inspect.cleandoc(func.__doc__) % self.params
        return func

    def update(self, *args, **kwargs):
        """
        Update ``self.params`` (which must be a dict) with the supplied args.
        """
        self.params.update(*args, **kwargs)


[...]

Issue #2858, opened 2023-11-16
load_project() is not working properly


mostafayossef:
**Describe the bug**
When I try to load_project, I get the following error:

```
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[2], line 2
      1 filename='sim_01.pnm'
----> 2 pr03=op.utils.Workspace.load_project('sim_01',filename)

File ~\AppData\Local\anaconda3\Lib\site-packages\openpnm\utils\_workspace.py:252, in Workspace.load_project(self, filename)
    225 r"""
    226 Loads a Project from the specified 'pnm' file
    227 
   (...)
    249 
    250 """
    251 with open(filename, 'rb') as f:
--> 252     proj = pickle.load(f)
    253     proj.settings.uuid = str(uuid4())
    254     proj.name = self._validate_name(proj.name)

ModuleNotFoundError: No module named 'openpnm.utils.Project'
```

**To Reproduce**
**code snippet:**
```
import openpnm as op
import matplotlib.pyplot as plt
import numpy as np
import scipy.stats as spst
op.visualization.set_mpl_style()
filename='sim_01.pnm'
pr03=op.utils.Workspace.load_project('sim_01',filename)
```

**Expected behavior**
It should load the project, so I can check the network

**Screenshots**

**Additional context**

jgostick:
Thanks for reporting this.  It is coming from the pickle attempting to look-up a setting on the phase object before the settings have been initialized.  Still working on a fix, but it's top priority, so within a few days.  We are already working on a patch for a matplotlib error, so we'll roll this into that PR too.

PR #2864, opened 2023-11-20
Creating a patch release

jgostick:
Two critical bugs: 
1. fix matplotlib import error #2863
2. fix pickle loading error #2858


jgostick:
looks like pypardiso got it fixed, but our builds were not "seeing" the latest versions...maybe they were using a cached package?

PR #2880, opened 2023-12-22
Add more `scipy` integrators to handle stiff problems

mkaguer:
Currently there is only one scipy integrator classes that uses RK45. I created a general integrator class that uses RK45 by default called ScipyIntegrator, this is the parent class. Then, I created child classes that inherit from this class but pass different solvers. I added two more solvers currently not available, BDF and LSODA. These solvers work for stiff problems.


ma-sadeghi (changes requested):
Nice, can you add some unit tests?

mkaguer:
> Nice, can you add some unit tests?

Yea, I'll do that this week when I get back to work!

mleot:
I also have motivation to add more integrators to OpenPNM. I can help out with writing some of these unit tests.

jgostick:
Hey @mkaguer, are we still interested in merging this PR?  I have refreshed it by merge the latest `dev` into it.  It now passes the non-mac tests.  Should we merge it?  Any change you could address @ma-sadeghi's request to add a test or two?

TKlokgieters:
@jgostick Any chance this can be merged? As far as I can tell the unit tests were committed in/with [a6cc209] .
I'd also be interested in solving stiff problems using the different integrator methods.

If anything is missing I'd be happy to help.

Issue #2922, opened 2024-07-03
from scipy.optimize.nonlin import TerminationCondition invalid with scipy v >= 1.14


hpc-dave:
**Description**
With the update to scipy 1.14, the Termination condition is in scipy.optimize.linalg is removed. Subsequently, import of _reactive_transport.py fails, leading to a failure of importing openpnm as is

**Reproduction**
Steps to reproduce the behavior, ideally a code snippet that only reproduces the bug:
```python
import openpnm as op
```

**Expected behavior**
Should import without error

**Quick fix**
A quick fix for compatibility could be as follows:
```python
# dirty hack for compatibility with scipy 1.14
try:
    from scipy.optimize.nonlin import TerminationCondition
except ImportError:
    class TerminationCondition:
        def __init__(self, f_tol: float, f_rtol: float, x_rtol: float, norm: float):
            self.f_tol = f_tol
            self.f_rtol = f_rtol
            self.x_rtol = x_rtol
            self.norm = norm

        def check(self, f_tol: float, f_rtol: float, x_rtol: float, norm: float) -> bool:
            return f_tol < self.f_tol and f_rtol < self.f_rtol and x_rtol < self.x_rtol and norm < self.norm

```

ma-sadeghi:
Thanks for the bug report. Would you be willing to submit a PR with your fix?

hpc-dave:
Sure! I looked a bit more into it and figured out that a better, yet still not so nice fix means to import from `scipy.optimize._nonlin` instead. The file still exists, however I do not know for how long. It might be worthwhile to consider a more durable alternative.
I submitted PR #2923

AilsaZeng:
Hi, did you work out how to solve this problem? I still cannot use the import porespy as ps.

hpc-dave:
I don't think that is related to this issue. Importing porespy works for me with numpy 2.0.0
Maybe you can open another issue in the porespy repository (https://github.com/PMEAL/porespy/issues) and explain your problem in more details there!

PR #2923, opened 2024-07-04
hotfix import TerminationCondition for scipy v>=1.14

hpc-dave:
**Issue**
With updating to  scipy 1.14, import from `scipy.optimize.nonlin` leads to an import error

**Temporary Fix**
Import from `scipy.optimize._nonlin` instead

**Additional**
The use of `TerminalCondition` should be reviewed. In my opinion, substituting the check done by `TerminalCondition` would reduce the dependencies of the project and would not decrease readability.

Related to #2922

### DIFF
--- a/openpnm/algorithms/_reactive_transport.py
+++ b/openpnm/algorithms/_reactive_transport.py
@@ -3,7 +3,7 @@ import sys
 
 import numpy as np
 from numpy.linalg import norm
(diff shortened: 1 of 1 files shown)


hpc-dave:
The failed checks seem to be related to other numpy issues. I'll leave the fix for now as it is.

jgostick:
Had to 'redo' this PR in #2931.

PR #2959, opened 2025-02-04
Fix IO issues for 'io.network_from_csv'

hpc-dave:
Here, two issues are addressed:
1) A DtypeWarning appears when using 'io.network_from_csv', that mixed types were encountered. Since specifying the dtype is not generally possible, the flag low_memory=False has to be set to avoid this warning by pandas.
2) For the columns with boolean values, pandas keeps the general Object type (dtype('O')), which can lead to issues later on, especially when trying to use the values directly. Especially nasty to debug when used in combination with topotools. This issue is fixed by adding an additional conversion step at the end of the function 'io.network_from_csv'

Note:
- pytest is currently failing due to updates in scipy 1.15,*, see discussions [here](https://github.com/scipy/scipy/issues/22236) and [here](https://github.com/pytest-dev/pytest/issues/10845#issuecomment-2577509473)

### DIFF
--- a/openpnm/io/_csv.py
+++ b/openpnm/io/_csv.py
@@ -1,7 +1,7 @@
 import re
 
 import numpy as np
-from pandas import read_table
+from pandas import read_csv
 
 from openpnm.io import _parse_filename
 from openpnm.io._pandas import network_to_pandas, project_to_pandas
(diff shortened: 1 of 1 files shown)


jgostick:

PR #2981, opened 2025-09-12
Fix NumPy positional-only argument usage

gerlero:
Fixes the error ```TypeError: reshape() got some positional-only arguments passed as keyword arguments: 'a'``` when passing the first argument as a keyword argument.


gerlero:
Related/similar: #2958

jgostick:

PR #2982, opened 2025-09-12
Fix chemical names not being recognized

gerlero:
Fix the errors like  ```ValueError: Chemical name (o2) not recognized``` (apparently caused by a change in how `thermo` handles names).


gerlero:
A bunch of the CI errors should be fixed by #2981. Unless you prefer otherwise, I'd rather we go through #2981 first and, then I can rebase this one (and also #2982 and #2983) on top of those changes.

jgostick:
I am working on a PR to add uv support, and I merged #2981 into that locally. I believe that your contributions will show up in the git history (thanks!), but I'll cancel that PR.

gerlero:
Sure, no problem for me with the commit authorship (and it'll also probably show correctly anyway).

jgostick:

PR #2984, opened 2025-09-12
Identify uninvaded throats between previously invaded pores within same cluster

gerlero:
Fix an issue with the `drainage` algorithm.


gerlero:
@jgostick Thanks! May I ask for a release with the newly merged changes?

PR #2985, opened 2025-09-13
Moving to uv

jgostick:


gerlero:
@jgostick Heads up, the `dev` group didn't work since it should go under `[dependency-groups]`, while it was under `[project.optional-dependencies]`

gerlero:
BTW if you want to refer to other _optional dependencies_ (not other _dependency groups_), you have to do:

```
dev = [
    "openpnm[test]",
    "openpnm[docs]",
]
```

instead of the other syntax.

FWIW, the difference between _optional dependencies_ (installable along with the package with `pip`) and _dependency groups_ (only for development) is kinda confusing; I ran into similar issues and had to learn it too.

PR #2986, opened 2025-09-14
Merging dev into release for V3.5.1

jgostick:


jgostick:
perhaps this release could be a patch: 3.5.1 since it doesn't have any new features?

PR #2989, opened 2025-09-17
Add uv.lock to .gitignore

gerlero:
Add uv.lock to .gitignore file, to prevent it from being inadvertently committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,9 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+# uv lockfiles
+uv.lock
+
 docs/_build/
 public/
 


jgostick:
Hi @gerlero, I am pretty new to uv, but isn't the lock file important so that one can know *exactly* which dependencies work? When is the lock file used by uv?  Doesn't it just use pyproject.toml?

gerlero:
@jgostick The rule of thumb is that the lockfile is usually committed if the project is an application (where everyone is expected to use the same versions of dependencies, including for deployment) and ignored if the project is a library (where users are expected to be using a wide range of compatible dependency versions). OpenPNM falls in the latter group, so I'd recommend ignoring it in this case.

uv only uses the lockfile when installing the repo for development (with `uv sync`, usually after a `git clone`); so you can't force non-developer users to install specific versions of dependencies via the lockfile (you can only use the `pyproject.toml` file for that).

In the end it's your call, but I haven't seen any libary project yet that commits its lockfile.

EDIT: Probably obvious, but just to be extra clear to anyone else reading, the lockfile has not been committed yet to this project. This change just tries to make sure that it doesn't get added by accident.

PR #2990, opened 2025-09-17
Install pypardiso by default on x86_64 platforms

gerlero:
As a suggestion only, conditionally install `pypardiso` by default on compatible platforms.


jgostick:
Make sense.  I had started thinking along those line in some of the actions, but neven followed through with all of them.

gerlero:
Unfortunately, it looks like this change is not working properly for `pip` installs of `openpnm`, since there's also a list of dependencies at `setup.py` which seems to be used for releases. 

While this could be fixed by added the same constraint in `setup.py`, I think I'll open a PR suggesting condensing everything packaging-related within `pyproject.toml`.

PR #2991, opened 2025-09-18
Fix mistake in pore network size consistency example 6

TKlokgieters:
The text and code (sometimes) assumed a 4x4x1 pore network, while a 5x5x1 is used in the example so I changed to example to consistently reference a 5x5x1 pore network.
Additionally I added a warning about overlapping subdomains


jgostick:
Thank you very much for your contribution!

PR #2998, opened 2025-10-07
Remove setup.py

gerlero:
Move missing stuff from `setup.py` to `pyproject.toml`, and remove the earlier.

Builds (e.g. those from the `publish-to-pypi.yml` workflows) now (expectedly) use the dependency list in `pyproject.toml`. I chose to use `python -m build` as the build frontend, but `uv build` would be equivalent (assuming uv is installed).


jgostick:
This looks great, and thanks so much for dealing with all this issues.

PR #2999, opened 2025-10-07
Hotfix to make the install/build process more robust

jgostick:


jgostick:
Hi @gerlero, you seem to be pretty handy with these build tasks....any idea why the python 3.13 builds are taking forever?  Googling is not giving me much luck.

gerlero:
@jgostick macOS might be expected due to pypardiso not being installed there, but I'm at a loss about the other ones. Can you run them again? (Curiously, the install part is also taking longer with those runs, so it might be some strange runner issue?)

jgostick:
This has been slow on all 3.13 builds since we added 3.13 as a option during the last spurt of activity.

gerlero:
@jgostick That's really strange. When installing with the `uv.lock` file, there are exactly zero differences in dependency versions (even of indirect dependencies) between Python 3.12 and 3.13 (I checked this manually on Linux). 

So, it must be either the interpreter or some compiled dependency that is causing it, which still would be strange.

PR #3001, opened 2025-10-14
Add integration method options to integrator class

TKlokgieters:
Sometimes I'd like to be able to have different integration methods for transient solution algorithms.
I changed the class to allow choosing a method and renamed it for clarity. 
The old class name (`ScipyRK45`) is kept to avoid any code breaks.

I'm not certain whether the documentation is currently correctly implemented such that for `ScipyRK45` it automatically copies the relevant parts from `ScipySolveIVP`.

Thanks for you consideration


TKlokgieters:
I noticed this is the same as suggested in #2880

PR #3004, opened 2025-10-20
Fix handling of trapped pores and throats in invasion sequence

gerlero:
Follow-up to/missing from #2983 and #2984


gerlero:
Hi @jgostick, sorry to bother! When you have a moment, could you take a look at this PR?
It would save us from having to patch the source ourselves just for this small change. Thanks!

PR #3026, opened 2025-11-21
Optimize  _find_trapped_pores() in invasion percolation trapping (~20× faster) [#enh]

kgrox:
Summary
------
The time complexity of the reverse invasion percolation procedure according to the reference should be nearly O(N), so the `apply_trapping()` should run at least as fast as the `run()`, which is O(NlogN). but when I was calculating a cubic network with a shape of [100,100,100], and a connectivity of 6, the `run()` cost around 4 seconds and the apply_trapping() cost more than 100 seconds.
So I checked the code and rewrite the _find_trapped_pores() function, and the union-find set in _qupc.py. the optimized code could now get the same result within 2 second. 

Performance
-----------
On the network I mentioned above:

- Before: 
Forward IP simulation time: 4.230068 seconds
Reverse IP simulation time: 103.905001 seconds
- After:
Forward IP simulation time: 4.319790 seconds
Reverse IP simulation time: 1.639167 seconds

the code i used to get these result:
```
import openpnm as op
import numpy as np
import time
import matplotlib.pyplot as plt

np.random.seed(12345)   
pn = op.network.Demo(shape=[100, 100, 100], spacing=1)
pn["throat.volume"] = 0
src_pore=[[-1,0,0]]
vis_conns = [[i, 1000000] for i in range(0, 100)]
op.topotools.extend(network=pn,coords=src_pore)
op.topotools.extend(network=pn,conns=vis_conns)
pn["pore.volume"][-1]=0.0
pn["throat.volume"]=0.0
hg = op.phase.Mercury(network=pn)
f = op.models.physics.capillary_pressure.washburn
hg.add_model(propname='throat.entry_pressure',
             model=f, 
             surface_tension='throat.surface_tension',
             contact_angle='throat.contact_angle',
             diameter='throat.diameter',)
hg["throat.entry_pressure"][-1:-101:-1] = hg["throat.entry_pressure"][1:-101].min()

inlet=[i for i in range(0,100)]
outlet=[i for i in range(pn.Np-101,pn.Np-1)]

mip = op.algorithms.InvasionPercolation(network=pn, phase=hg)

mip.set_inlet_BC(pores=inlet)
mip.set_outlet_BC(pores=outlet) 

start_time = time.perf_counter()
mip.run()
end_time = time.perf_counter()

print(f"Forward IP simulation time: {end_time - start_time:.6f} seconds")

start_time = time.perf_counter()
mip.apply_trapping()
end_time = time.perf_counter()

print(f"Reverse IP simulation time: {end_time - start_time:.6f} seconds")

data=mip.pc_curve()
np.savetxt("ip_pc_curve.txt",data)
```
I've check the result to make sure the optimized code get the exact same result

### DIFF
--- a/openpnm/_skgraph/queries/_qupc.py
+++ b/openpnm/_skgraph/queries/_qupc.py
@@ -8,6 +8,8 @@ __all__ = [
     'qupc_update',
     'qupc_compress',
     'qupc_reduce',
+    'qupc_union',
+    'qupc_find',
 ]
 
 
@@ -41,6 +43,20 @@ def qupc_reduce(arr):
     for i in range(len(arr)-1, 0, -1):
         arr[i] = arr[arr[i]]
     return arr
+
+@njit
+def qupc_union(arr, ind, val):
+    indp=qupc_find(arr, ind)
+    valp=qupc_find(arr, val)
+    arr[indp]=valp
+    return arr
+
+@njit
+def qupc_find(arr, ind):
+    while arr[ind] != ind:
+        arr[ind] = arr[arr[ind]]
+        ind = arr[ind]
+    return ind
 
 
 if __name__ == '__main__':
--- a/openpnm/algorithms/_invasion_percolation.py
+++ b/openpnm/algorithms/_invasion_percolation.py
@@ -6,7 +6,7 @@ import numpy as np
 from numba import jit, njit
 from tqdm.auto import tqdm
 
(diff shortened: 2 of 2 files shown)


jgostick:
Thank you very, very much for this!  Our team has shrunk a lot, and I'm really just working on my own to maintain the nuts-and-bolts of this package. This contribution is really timely and appreciated.

Issue #3042, opened 2026-04-15
Numpy reshape change in variable name


TKlokgieters:
From numpy 2.1 in the function `np.reshape()` the variable `newshape` is replaced by `shape`
And in Numpy 2.4 this change is definitive
#api

PR #3043, opened 2026-04-15
Use positional argument name for np.reshape: newshape. 

TKlokgieters:
in Numpy 2.1 the variable is changed from newshape into shape and from Numpy 2.4 the variable newshape is expired.
#api

#3042

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "matplotlib-inline>=0.1.7",
     "networkx>=3.5",
     "numba>=0.61",
-    "numpy>=2.1,<2.2",
+    "numpy<2.2",
     "pandas>=2.3.2",
     "pyamg>=5.3.0",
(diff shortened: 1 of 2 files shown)


TKlokgieters:
It is possible that a higher version than Numpy 2.2 is allowed, but I do not know how to check for it.

gerlero:
I'd suggest simply using the argument positionally as:

```python
np.reshape(temp, [len(edges), 2], order="F")
```

That way it will work regardless of the NumPy version.

TKlokgieters:
Thanks @gerlero that is indeed a cleaner solution. I changed the proposal to follow your suggestion.

jgostick:
Does anyone know why github only ran 2 checks, and none of the tests?  It is not showing me an option to 'allow tests' or anything.  Must be a setting somewhere?

jgostick:
Although, since only one line was changed, I'd say it might be safe to skip the tests.

jgostick:
Interestingly, the tests did run, as they are visible in the Actions tab, but the results are just not being reported in this thread.  It seems one of the example runs did not pass.  I will fix that before merging into release.

PR #3047, opened 2026-04-28
Update numpy version constraint in pyproject.toml

ecederstrand:
I'd like to use newer versions of numpy  with OpenPNM


ma-sadeghi:
@ecederstrand Thanks for the contribution!

